PR #356, opened 2015-08-23
Add ZYI to airports.dat

bprosnitz:


jpatokal:
Pull requests for *.dat files are not accepted, since they are snapshots of the live data, where you will find ZYI already. (And yes, a refresh of the snapshots is overdue.  Soon...)

PR #357, opened 2015-08-24
Added IATA code for Iwakuni Airport.

cdr6275:


jpatokal:
Pull requests for *.dat files are not accepted, since they are snapshots of the live data.  I have updated IWK in the database. (And yes, a refresh of the snapshots is overdue.  Soon...)

PR #383, opened 2015-10-09
Set "Queenstown International" as the airport name

asfaltboy:
Rather then the New Zealand city, which is simply "Queenstown".


jpatokal:

(Although this should be fixed already?)

PR #405, opened 2015-11-09
Update encoding information

lehmannro:
It seems the documentation on exports' encodings is outdated and all those files are now in UTF-8.  I took "with no special characters" to mean non-ASCII (which is only true for the routes data, and not airports/airlines.)

### DIFF
--- a/data.html
+++ b/data.html
@@ -83,7 +83,7 @@
 	      </tr>
 	    </table>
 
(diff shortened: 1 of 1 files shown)


jpatokal:
Sorry for the wait!  This patch is no longer compatible, but I've made the fix manually.

Issue #414, opened 2015-11-30
China Lake Naws 


chadliu23:
The location should be "China Lake", not "China"

chadliu23:
see merge request #415

chadliu23:
http://www.airnav.com/airport/NID

jpatokal:
Updated, thank you!

chadliu23:
When you will update latest airport to github?

PR #415, opened 2015-11-30
#414 Change location of "China lake Naws"

chadliu23:
from China to China Lake


jpatokal:
Sorry: airport data is extracts from the live database, so pull requests are not accepted.

PR #422, opened 2015-12-06
Update messages.po

Madeco:
Updated some translations


jpatokal:
Whoah, I completely missed this earlier.  Thanks for the fixes!

PR #425, opened 2015-12-07
airports: renamed ICAO code for Lima airport from SPIM to SPJC

aleksander0m:
Since November 12th 2015, the Jorge Chávez International Airport uses the new
"SPJC" ICAO code, instead of the previous "SPIM".

https://en.wikipedia.org/wiki/Jorge_Ch%C3%A1vez_International_Airport


jpatokal:

Although this is probably fixed already?

PR #537, opened 2016-05-26
Fixed name of the Vadsø Airport

roalonso:
Fixed the entry for the Vadsø Airport in `airports.dat`.


jpatokal:

Please submit your fix upstream to OurAirports if you're still seeing this, although this is likely fixed already?

PR #562, opened 2016-07-06
Fix Deer Valley Municipal Airport location and timezone

damsonn:
Was pointing to somewhere in China...


jpatokal:

Please submit your fix upstream to OurAirports if you're still seeing this, although this is likely fixed already?

Issue #682, opened 2017-02-10
Update airport Doha Intl (DOH/OTDB)


jpatokal:
New airport edit suggestion submitted by Poderemergente2050:

INSERT INTO airports(name,city,country,iata,icao,x,y,elevation,timezone,dst,uid) VALUES('Doha Intl', 'Doha', 'Qatar', 'DOH', 'OTDB', 51.565056, 25.261125, 35,  3, 'U', 34682);

Existing, potentially conflicting airport information:

```
Array
(
    [name] => Hamad Intl
    [city] => Doha
    [country] => Qatar
    [iata] => DOH
    [icao] => OTHH
    [x] => 51.608056
    [y] => 25.273056
    [elevation] => 35
    [apid] => 11051
    [uid] => 3
    [timezone] => 3
    [dst] => N
    [tz_id] => 
    [type] => unknown
    [source] => User
)

```

Cross-check this edit on other sites with compatible licensing:
- OurAirports: http://ourairports.com/airports/OTDB/pilot-info.html
- Wikipedia: http://www.google.com/search?q=wikipedia%20OTDB%20airport&btnI

PR #692, opened 2017-03-10
Add Doha International Airport

Jakuje:
Source:
https://en.wikipedia.org/wiki/Doha_International_Airport


Jakuje:
oh .. I see it is already tracked in the #682

jpatokal:
Please submit any airport data changes upstream to http://ourairports.com/.

PR #736, opened 2017-05-27
modif_data

davebulaval:
Correction of the  tzFormat variable for all the airports. 
We initially only wanted to fill missing values, but we then realize that actual information was sometime wrong or ambiguous.
Used sp and rgdal packages to extract geospatial information from the shapefiles with R. Plus we have used tz_worlwide for cross validation.

[timezone wiki source](https://en.wikipedia.org/wiki/Tz_database)
[Source of the dataset](http://web.cs.ucla.edu/~eggert/tz/tz-link.htm)
[shapefile](https://github.com/evansiroky/timezone-boundary-builder)
[Tz_worldiw shapefile (for cross validation)](http://efele.net/maps/tz/world/)

About the commit;Modification of tzFormat:
In modification file:
- source code for modification and verification
- website source
- visual verification of some airports


SamuelCabralCruz:
## Precisions
- Tz_worldiw stands for **tz_world**
- tzFormat is the name that we use to represent the **Tz database time zone** variable in the airports.dat table
- The **_errors.md_** file is a log of all the known tzFormat that were inconsistent with the merged time zones from **tz_world** or **timezone-boundary-builder** shapefiles.
- The **_valid.md_** file is a log of all the merged time zones that were in discrepancy.
- The **_.png_** files correspond to screenshots taken from Google Maps.
- After a quick analysis of the **_valid.md_**, we realized that the **timezone-boundary-builder** shapefile renders better results in almost all the cases (only 2 out of 22 airports were better represented by the **tz_world** shapefile and in only 1 case the **tzFormat** was correct).

## Warnings 
- The dataset still have few **missing values** (7/7184 records)
- The timezone **America/Punta_Arenas** is not recognized by **lubridate** package as a valid timezone
- However, **tzFormat** was mainly setting all those airports in the **America/Santiago** timezone which may introduce errors in the flight time calculation ([See](https://www.timeanddate.com/time/map/#!cities=574))
- Only 5 airports are currently assigned to **America/Punta_Arenas**
- No work has been done over the timezone related variables, those variables should be extrapolated from the new information or simply deleted since the timezone format ID contains everything we have to know.

jpatokal:
Thanks for the effort, but I really wish you had contacted me first, because I'm unable to accept this for multiple reasons:
- airports.dat is periodically regenerated from live data, so I don't accept diffs for it
- airports.dat file is completely mangled, it's supposed to be a CSV
- the code you have written is a one-off update, I need a repeatable process and one already exists at https://github.com/jpatokal/openflights/blob/master/data/update-timezones.py

PR #804, opened 2017-11-22
EDAL has not been operational for a while now. Removed entry id 8500

llinear-zz:
Just removed an outdated entry in the airport database


jpatokal:
Old airports can't be deleted, because that would break any flight data linked to them.  Also, the airport data is periodically regenerated from OurAirports, meaning pull requests aren't accepted anyway: https://github.com/jpatokal/openflights/blob/master/data/README.md

Issue #825, opened 2018-01-06
FlightMemory import--truncation of note field


dahorowitz:
When I attempt to import the html file saved from my FlightMemory account, the notes field is truncated for every entry--for example, the note "Boarding position B35" in the html file is truncated to "position B35" in the import process, and "With Jordan/7A" becomes "an/7A". It appears that the first nine characters of every note are being ignored.

Tmpfile phpllrwJL
[page1.html.txt](https://github.com/jpatokal/openflights/files/1608936/page1.html.txt)

jpatokal:
New airport edit suggestion submitted by AnnaZu:

UPDATE airports SET name='Barajas', city='Madrid', country='Spain', iata=NULL, icao=NULL, x=-3.593554, y=40.491761, elevation=2027, timezone=1, dst='E' WHERE apid=10599;

Existing, potentially conflicting airport information:

```
Array
(
    [name] => Aeropuerto T4 Madrid Metro
    [city] => Madrid
    [country] => Spain
    [iata] => 
    [icao] => 
    [x] => -3.593554
    [y] => 40.491761
    [elevation] => 2027
    [apid] => 10599
    [uid] => 5329
    [timezone] => 1
    [dst] => E
    [tz_id] => 
    [type] => unknown
    [source] => User
)

```

Cross-check this edit on other sites with compatible licensing:
- OurAirports: http://ourairports.com/airports//pilot-info.html
- Wikipedia: http://www.google.com/search?q=wikipedia%20%20airport&btnI

arcykong:
FlightMemory format has been changed a little bit,
From: `<td class="liste_rot"><span title="Comment: 2.5-hr delay due to tire puncture">Com</span><br> ...`
To: `<td class="liste_rot"><span title="2.5-hr delay due to tire puncture">Com</span><br> ...`

I have added checking routine trim comment only when "Comment: " exists.

jpatokal:
@arcykong Thanks for the fix, it's live now!

PR #833, opened 2018-02-21
Fix FlightMemory comment import (#825)

arcykong:
Trim comment when only "Comment: " exists.


jpatokal:
A test for this in test/server/import.php would be nice, but I'll take what I can get -- thanks!

PR #834, opened 2018-02-24
Updated Loganair to add IATA-code and set it as active

backeby:
Updated information for airline Loganair


jpatokal:

PR #861, opened 2018-06-15
Refactor test suite to use PDO

williammck:
This is part of a larger refactor to move to using PDO globally, moving away from `mysql_*`, and fix several compatibility issues with PHP 7.

Nothing's changed in the way of functionality here -- just plugging away at converting everything to our new friend PDO.


jpatokal:
Thank you, this is great!

PR #928, opened 2019-01-16
Region De Murcia Airport Added

data-runner:
I've updated the file to include the new Region de Murcia International Airport. I've included all the details I could find and checked this against FlightAware.


data-runner:
Just discovered it has an elevation of 633. Will add that and update

data-runner:
updated!

PR #938, opened 2019-03-03
Refactor website to use PDO

williammck:
Sorry this took so long, last year got a bit hectic!

This PR is somewhat large - I've split it up into several different commits to make it more easily digestible.

- The first two commits are tied to the UI, one dealing with locales and the other just general changes from `mysql_` to `PDO` in the UI views.
- The next commit is all the backend work. This is the big one (oops, sorry).
   - This also makes some changes to how Github issues are generated (in `php/apsearch.php`); let me know if you have any feedback on how they should look.
- The last three commits are dealing with some leftover cruft - fixing tests and doing a pass on `getFilterString` to make sure everything gets escaped correctly.

After all this work was done, I couldn't find any remaining references to `php/db.php`, but left it in there just in case.

All tests pass for me, but please go ahead and try whatever you want against it.

### DIFF
--- a/badge/banner.php
+++ b/badge/banner.php
@@ -1,5 +1,5 @@
 <?php
-include '../php/db.php';
+include '../php/db_pdo.php';
 include '../php/helper.php';
 
 putenv('GDFONTPATH=' . realpath('.'));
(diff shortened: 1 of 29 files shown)


jpatokal:
Whoah, thanks so much for this!  I've just pushed through some changes that need to settle in first, but I'll try to merge this soon.

PR #954, opened 2019-05-06
Update planes.dat

SteveWaring:
Add missing planes from Wikipedia. Planes from manufacturers not already in the file have been excluded. Contradictions between existing lines in the file and Wikipedia have been left unchanged.


data/planes.dat
```diff
@@ -104,24 +137,43 @@
...
+"Douglas DC-8-50","D8T","DC85"
+"Douglas DC-8-62","D8L","DC86"
+"Douglas DC-8-72","D8Q","DC87"
 "Douglas DC-9-10","D91","DC91"
+"Douglas DC-9-20","D92","DC92"
 "Douglas DC-9-30","D93","DC93"
 "Douglas DC-9-40","D94","DC94"
 "Douglas DC-9-50","D95","DC95"
 "Embraer 170","E70","E170"
+"Embraer 175","E75","E75S/L"
```
jpatokal: No slashes in code please, just set the ICAO code here to null.

jpatokal (changes requested):
Thanks for creating this!  One thing to fix before checking in.

PR #992, opened 2019-09-25
Added distinct plane type count

ssegraves:


php/stats.php
```diff
@@ -74,7 +74,7 @@
 }
 
 // unique airlines, unique planes, total distance (mi), average distance (localized), average duration
```
reedy, line 76: Comment presumably wants updating to match...

reedy: And because GitHub CR is daft, I can't leave this comment at line 84 below... But presumably this doesn't actually result in anything being output, on the website, it just makes it available to JS, so would also need a change to https://github.com/jpatokal/openflights/blob/89643b0b3a6d750459c40799b45901beb080d805/openflights.js#L1558-L1564 too, to actually display the output?

Obviously doesn't block it being merged, but the comment should probably be updated.

ssegraves: I'll make the necessary changes and submit another pull. Thanks @reedy

chrisrosset:
This PR seems dead but, nevertheless, I don't think this change is right.

Most importantly, the website already shows the plane type count (Analyze / Unique / Vehicles). This information is collected on the line that's being edited:

```sql
COUNT(DISTINCT plid) AS num_planes
```

The `plid` field ("Plane ID") [is the primary key for the planes table](https://github.com/jpatokal/openflights/blob/cc1a3b460b6654d59f3075bd977c8c68f3e7c36b/sql/create.sql#L163) and uniquely identifies a plane type.

Individual planes are not catalogued by the system but can be tracked by the user. These would be identified by the [`registration` column in the flights table](https://github.com/jpatokal/openflights/blob/cc1a3b460b6654d59f3075bd977c8c68f3e7c36b/sql/create.sql#L134).

I'm not aware of IATA assigning any codes to plane types. They do assign [airline and airport codes](https://www.iata.org/en/publications/directories/code-search/) and the database stores these codes in the `airlines` and `airports` tables respectively.

reedy:
https://en.wikipedia.org/wiki/List_of_aircraft_type_designators

image

ssegraves:
Maybe this change needs a little more design behind it. I will take some time to think about what all would need to be changed overall to make the unique plane identifier available and used in the count. I wish I would have put more comments in my original PR. 

To @chrisrosset's point, it would be helpful to have a count by registrations as well.

chrisrosset:
@reedy, thank you for the link! Today I learned.

~I took another look at the `planes` table I think we might already have this data in the `abbr` column.~ Nevermind, the DB has both IATA and ICAO codes which I discovered while doing #1210 and #1211.

@ssegraves Could you please take a look at https://github.com/jpatokal/openflights/blob/master/data/planes.dat and double-check if the middle column (e.g. `321` in `"Airbus A321","321","A321"`) matches your expectations?

I think you'd need to left join with `planes` on `plid` to have access to them in the query.

2sylbl:
The issue I see with this is that currently users can specify their own planes. For example, planes.dat only includes CRJs as "Canadair Regional Jet." I have standardized on calling them all "Bombardier CRJ-xxx" in my account, since they have long been branded as that. (Technically, they should become Mitsubishi at this point I guess). So I don't think that my entries are going to match up with the IATA/ICAO codes in planes.dat unless I'm missing something. 

Incidentally, I have an outstanding MR for planes.dat to add the COMAC ARJ-21 - maybe one of you guys can merge that in? https://github.com/jpatokal/openflights/pull/1025/commits

PR #1029, opened 2020-05-07
Updated Alaska Airlines callsign and country value

fetchy-cha:
Entry 439, Alaska Airlines, Changed:
`callsign`: " Inc." -> "ALASKA"
`country`: "ALASKA" -> "United States"

I found the problem while pulling airline data from https://openflights.org/html/alsearch


dlenski:
Also fixed this in #1038

PR #1038, opened 2020-10-19
Fix two airlines

dlenski:
- Alaska airlines should have `United States` as country
- Formosa Airlines ceased operating decades ago (https://en.wikipedia.org/wiki/Formosa_Airlines) and its IATA code (`LY`) is now used by Vueling.


luisjoujr:
When introducing a VY flight number, still appears the defunct Formosa Airlines (VY) as default. Now it should read Vueling Airlines (VY).

PR #1098, opened 2022-10-16
composer.json: Move test dependancies to require-dev

reedy:


reedy:
Merged into https://github.com/jpatokal/openflights/pull/1084

PR #1135, opened 2022-12-05
flights.php: match %s count to parameter count

reedy:


jpatokal:
Thanks for the fix, I'm mystified how this was not spotted earlier.  Apparently PHP used to be really permissive.

reedy:
> Thanks for the fix, I'm mystified how this was not spotted earlier. Apparently PHP used to be really permissive.

It still is in many many places...

Issue #1085, opened 2022-03-24
Software Licensing


reedy:
What license is this software? There is no COPYING or similar...

jpatokal:
OpenFlights is (supposed to be) AGPL and I'm rather astonished to realize that this was not obvious on Github.  Added a LICENSE file.

PR #1184, opened 2023-02-28
composer.json: Expand metadata

reedy:
Follows up https://github.com/jpatokal/openflights/issues/1085

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,16 @@
 {
+    "name": "jpatokal/openflights",
+    "description": "Website for storing flight information, rendering paths on a zoomable world map and calculating statistics, with plenty of free airline, airport and route data",
+    "type": "project",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,4 +1,16 @@
 {
+    "name": "jpatokal/openflights",
+    "description": "Website for storing flight information, rendering paths on a zoomable world map and calculating statistics, with plenty of free airline, airport and route data",
+    "type": "project",
+    "keywords": [
+        "openflights"
+    ],
+    "license": "AGPL-3.0-only",
```
reedy, line 8: @jpatokal Is this right? Or should it be `"AGPL-3.0-or-later"`?

reedy:
If it's indeed "or-later" that can be fixed later...!

PR #1206, opened 2023-06-17
Development Docker Setup

chrisrosset:
This tries to automate most of the setup instructions present in the `INSTALL` file by using Docker. I don't know if this is a good way of getting this up and running locally but it worked for me so I wanted to share it.

It's a bit janky in places - I know. I'd be happy to improve it if there's any feedback.

If it's not useful, no problem.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM ubuntu:22.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+ENV TZ=Etc/UTC
+
+RUN apt-get update && apt-get install -y \
+    apt-transport-https \
+    curl \
+    git \
+    locales \
+    mysql-client \
+    nginx \
+    software-properties-common \
+    unzip \
+    tzdata \
+    vim
+
+# System-level PHP setup
+RUN add-apt-repository ppa:ondrej/php -y
+RUN apt-get install -y \
+    php7.4 php7.4-curl php7.4-fpm php7.4-gd php7.4-mbstring php7.4-mysql php7.4-xml
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
+RUN sed -i -e '/catch_workers_output =/s/.*/catch_workers_output = yes/' /etc/php/7.4/fpm/pool.d/www.conf
+
+# nginx setup
+# note that these symlinks require the repo be mounted @ /var/www/openflights
+RUN rm /etc/nginx/sites-enabled/default
+RUN ln -s /var/www/openflights/nginx/openflights-dev /etc/nginx/sites-available/
+RUN ln -s /etc/nginx/sites-available/openflights-dev /etc/nginx/sites-enabled/
+
+CMD ["/bin/bash", "-c", "/etc/init.d/php7.4-fpm start; /etc/init.d/nginx start; tail -f /var/log/php7.4-fpm.log"]
--- a/README.md
+++ b/README.md
@@ -27,3 +27,21 @@ Test coverage is woefully incomplete, but comes in three flavors:
 ## Installation
 
 See [INSTALL](INSTALL) for system requirements and instructions.
+
+# Development Docker
+
+A basic Docker setup is available to simplify setting up a development environment but requires a
+couple of manual steps to get the site up and running.
+
+1. `cp php/config.sample.php php/config.php` and set `$host = "db";` so the host name matches the
+   the database container host name.
+2. Run `docker-compose up` to create the containers.
+3. Install local PHP dependencies inside the container.
+   ```
+   # host shell
+   user@host:openflights $ docker exec -it openflights-web-1 bash
+   
+   # container shell
+   root@ee261e8f9103:/# cd /var/www/openflights/ && php /usr/local/bin/composer install
+   ```
+4. You should be able to access the site at `http://localhost:8007`.
--- a/nginx/openflights-dev
+++ b/nginx/openflights-dev
@@ -1,0 +1,39 @@
+server {
+    root /var/www/openflights;
+
+    index index.html index.php;
+
+    location / {
+        ssi on;
+        if (-e $request_filename.php){
+            rewrite ^(.*)$ /$1.php;
+        }
+        try_files $uri $uri.html $uri/ =404;
+    }
+
+    location ~ \.php$ {
+        include snippets/fastcgi-php.conf;
+        fastcgi_pass unix:/var/run/php/php7.4-fpm.sock;
+        fastcgi_buffers 16 32k;
+        fastcgi_buffer_size 64k;
+        fastcgi_busy_buffers_size 64k;
+    }
+
+    location /banner/ {
+        rewrite "^/banner/(.+).png$" /badge/banner.php?user=$1;
+    }
+
+    if ($uri !~ \.) {
+        rewrite "^/airline/(.+)$" /index.php?query=$1;
+        rewrite "^/airport/(.+)$" /index.php?query=$1;
+        rewrite "^/blog/(.+)$" https://blog.openflights.org/$1;
+        rewrite "^/trip/(.+)$" /index.php?trid=$1;
+        rewrite "^/user/(.+)$" /index.php?user=$1;
+    }
+
+    location ~ /\.ht {
+        deny all;
+    }
+
+    listen 80 default_server;
+}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,29 @@
+version: '3.8'
+services:
+  db:
+    image: mysql:8.0
+    cap_add:
+      - SYS_NICE
+    restart: always
+    environment:
+      - MYSQL_ALLOW_EMPTY_PASSWORD=yes
+    ports:
+      - '3306:3306'
+    volumes:
+      - db:/var/lib/mysql
+      - .:/openflights
+      - ./sql/create.sql:/docker-entrypoint-initdb.d/1.sql
+      - ./sql/docker-seed.sh:/docker-entrypoint-initdb.d/2.sh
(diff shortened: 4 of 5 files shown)


README.md
```diff
@@ -26,3 +26,26 @@ Test coverage is woefully incomplete, but comes in three flavors:
...
+1. `cp php/config.sample.php php/config.php` and set `$host = "db";` so the host name matches the
+   the database container host name.
+1. Update `sql/create.sql` to allow database access across containers.
+   ```sql
+    CREATE USER openflights@'172.16.0.0/255.240.0.0';
+    GRANT ALL PRIVILEGES ON flightdb2.* TO openflights@'172.16.0.0/255.240.0.0';
+   ```
+
+1. Run `docker-compose up` to create the containers.
+1. Install local PHP dependencies inside the container.
```
reedy: Is this the diff being weird... Or are all of these `1.`?

reedy:
Thanks!

I know @jpatokal was looking to do something like this...

I'm not sure offhand how many things are hardcoded `https` in the code, so whether this may result in broken links if nginx is only listening on port 80 (I know there's fun and games with self signed certs too...)

README.md
chrisrosset: They are all `1.`. Markdown doesn't care about actual number when rendering and it makes editing easier since you can re-order without having to re-number. I'll change these to be numbered properly to make reading the plaintext easier.

chrisrosset:
> Thanks!
> 
> I know @jpatokal was looking to do something like this...
> 
> I'm not sure offhand how many things are hardcoded `https` in the code, so whether this may result in broken links if nginx is only listening on port 80 (I know there's fun and games with self signed certs too...)

Having spent a bit more time with this setup, I've already found a couple of `https` only links (e.g. export buttons). I'll have to do some more work to enable `https`.

Still, I think even in it's current form it's already useful as most of the functionality is available. As a side note, I've been using another container with PHP 8 (because the INSTALL notes mention that it's not compatible) installed to manually test side-by-side. I actually haven't found any breakage on that front (though there are quite a few warnings).

reedy:
I've been fixing code up that I "knew" should be problematic on PHP 8...

If you've got warnings, feel free to dump them on another issue and I'll take a look. A lot of them should relatively easily resolved.

reedy:
> Still, I think even in it's current form it's already useful as most of the functionality is available. As a side note, I've been using another container with PHP 8 (because the INSTALL notes mention that it's not compatible) installed to manually test side-by-side. I actually haven't found any breakage on that front (though there are quite a few warnings).

FWIW, I would merge this without it being "perfect"; I would want to at least quickly test it first though ;)

README.md
```diff
@@ -26,3 +26,21 @@ Test coverage is woefully incomplete, but comes in three flavors:
...
+
+# Development Docker
+
+A basic Docker setup is available to simplify setting up a development environment but requires a
+couple of manual steps to get the site up and running.
+
+1. `cp php/config.sample.php php/config.php` and set `$host = "db";` so the host name matches the
+   the database container host name.
+2. Run `docker-compose up` to create the containers.
+3. Install local PHP dependencies inside the container.
```
reedy, line 39: Any reason we can't run composer automatically?

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    php7.4 php7.4-curl php7.4-fpm php7.4-gd php7.4-mbstring php7.4-mysql php7.4-xml
+RUN mkdir /run/php # For the php-fpm socket file
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
+RUN sed -i -e '/catch_workers_output =/s/.*/catch_workers_output = yes/' /etc/php/7.4/fpm/pool.d/www.conf
+
+# nginx setup
+# note that these symlinks require the repo be mounted @ /var/www/openflights
+RUN rm /etc/nginx/sites-enabled/default
+RUN ln -s /var/www/openflights/nginx/openflights-dev /etc/nginx/sites-available/
+RUN ln -s /var/www/openflights/nginx/openflights-dev /etc/nginx/sites-enabled/
```
reedy: Isn't it more usual to chain the `sites-enabled/foo` to symlink to `sites-available/foo`, even if that is a symlink itself?

reedy (commented):
I'm not going to claim to be any sort of Docker expert (to any extent), but it should(TM) be possible to make a common "base" image, and then have the differences between the images (pretty much the PHP install stuff at this point)

https://docs.docker.com/build/building/multi-stage/ as at least one way

I don't know if this is any harder/easier with docker compose too...

It probably doesn't make too much difference as things are relatively simple, and probably aren't be going to get muc more complex in future, but it would reduce some duplication.

README.md
chrisrosset: We only mount the source code when the container is created. It might be possible as part of `CMD` but this wasn't a priority for me. If you have PHP installed locally, you can run it outside of Docker too.

Dockerfile
chrisrosset: /shrug

We don't even need the `sites-available` one.

chrisrosset:
> It probably doesn't make too much difference as things are relatively simple, and probably aren't be going to get muc more complex in future, but it would reduce some duplication.

Wrapping up the PHP8 migration would achieve the same de-duplication. :)

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    software-properties-common \
+    unzip \
+    tzdata \
+    vim
+
+# System-level PHP setup
+RUN add-apt-repository ppa:ondrej/php -y
+RUN apt-get install -y \
+    php7.4 php7.4-curl php7.4-fpm php7.4-gd php7.4-mbstring php7.4-mysql php7.4-xml
+RUN mkdir /run/php # For the php-fpm socket file
```
reedy: `/var/run/php/` is used in the nginx config below...

And does this actually need doing manually?

Dockerfile-php8
```diff
@@ -0,0 +1,30 @@
...
+    nginx \
+    software-properties-common \
+    unzip \
+    tzdata \
+    vim
+
+# System-level PHP setup
+RUN apt-get install -y \
+    php8.1 php8.1-curl php8.1-fpm php8.1-gd php8.1-mbstring php8.1-mysql php8.1-xml
+RUN mkdir /run/php # For the php-fpm socket file
```
reedy: `/var/run/php/` is used in the nginx config below...

And does this actually need doing manually?

nginx/openflights-dev
```diff
@@ -0,0 +1,39 @@
...
+        ssi on;
+        if (-e $request_filename.php){
+            rewrite ^(.*)$ /$1.php;
+        }
+        try_files $uri $uri.html $uri/ =404;
+    }
+
+    location ~ \.php$ {
+        include snippets/fastcgi-php.conf;
+        fastcgi_pass unix:/var/run/php/php7.4-fpm.sock;
```
reedy, line 16: This isn't going to work for PHP 8.1 of course

reedy:
> > It probably doesn't make too much difference as things are relatively simple, and probably aren't be going to get muc more complex in future, but it would reduce some duplication.
> 
> Wrapping up the PHP8 migration would achieve the same de-duplication. :)

Also requires Jani having the time ;). Not as time intensive as rebuilding on a new machine, though. Mostly just installing the newer packages, minor change to the default nginx config, update and reboot...

But then there'll be later PHP upgrades to do. PHP 8.2 and PHP 8.3 will bring their own pain, almost certainly.

nginx/openflights-dev
chrisrosset: Right, forgot I have a local change for this in the PHP8 container.

I could symlink these to an unversioned name or update `/etc/php/fpm/pool.d/www.conf` to remove the version from the name.

Dockerfile-php8
chrisrosset: I don't remember the details - I did need it at some point.

chrisrosset:
How about I just remove the PHP8 one for now. It's a later addition which I was using for one of the PRs and definitely not needed in the MVP.

Dockerfile-php8
chrisrosset: It appears that `/var/run` is a symlink to `/run` on Ubuntu.

```
root@af5f1b02b4f1:/# ls -lah /var/run
lrwxrwxrwx 1 root root 4 Nov 30  2022 /var/run -> /run
```

Still need to double-check if this needs to be done manually.

chrisrosset: > `/var/run/php/` is used in the nginx config below...
> 
> And does this actually need doing manually?

OK, it's because we run `php-fpm` directly instead of going through `systemd`...

jpatokal:
PHP8 broke a *lot* of stuff, which is why I fell back on 7, and even that caused significant pain.  I'm hesitant to upgrade unless there's a pressing reason to do so.

Re: hardcoded https, these are messy leftovers from the days when OpenFlights was upgrading to be 100% HTTPS, but it's been HTTPS-only for years now so explicitly specifying the protocol should be unnecessary.

reedy:
FWIW, I've changed my dev install to using PHP 8.2, and while there's a bit of logspam (filing issues as I've seen them), it's mostly working just fine. I think we've fixed up most of the big issues.

PHP 7.4 has been EOL since late 2022 as per https://www.php.net/supported-versions.php

jpatokal:
OK, that's good to hear.  I do recall 8 upgrading a lot of warnings to errors, so I'm hoping these have been actually fixed at source as opposed to us just having different php.ini settings or something!

reedy:
I'm not quite confident to say it's time to upgrade now, but I think it should be in a near term plan

chrisrosset:
@reedy A couple of small updates if you want to test. Otherwise, I'll merge and we can improve upon this later.

Issue #1155, opened 2022-12-23
PHP Notice: Undefined index: fid in flights.php on line 29


reedy:
```
  $fid = $_POST["fid"];
```

What should it default to?

chrisrosset:
I looked a bit at this yesterday because I was hoping to eliminate PHP 7 warnings before looking at PHP 8 compatibility.

My understanding is that this code (all the reads out of `$_POST` and `$_GET`) were written assuming that reading an undefined returns `null`. For `fid`, I think this is corroborated by checks like `if ($fid && $fid != "0")` and `(!$fid || $fid == "0")`.

Is it a good idea to add `?? null` to all of them? Maybe it'd be better to declare all of the variables up top and assign the `null` default and then only re-assign if the value is present?

Or maybe we should just switch on the method (we seem to only have `GET` and `POST`) and just declare/assign all of them?

```php
switch($_SERVER['REQUEST_METHOD']) {
    case "POST":
        $apid = $_POST["id"] ?? null;
        $trid = $_POST["trid"] ?? null;
        $alid = $_POST["alid"] ?? null;
        $fid = $_POST["fid"] ?? null;
        $user = $_POST["user"] ?? null;
        $year = $_POST["year"] ?? null;
        break;
    case "GET":
        $apid = $_GET["id"] ?? null;
        $trid = $_GET["trid"] ?? null;
        $alid = $_GET["alid"] ?? null;
        $fid = $_GET["fid"] ?? null;
        $user = $_GET["user"] ?? null;
        $year = $_GET["year"] ?? null;
        break;
    default:
        die("Unsupported method");
}
```

reedy:
It's not worth declaring variables that aren't used; it just adds to confusion that you don't know if it's supposed to be used (and isn't) etc.

Most can probably be defaulted to null, when read though. This does seem to be common for the POST/GET variables, but not so consistent for other variables, for example like seen in https://github.com/jpatokal/openflights/pull/1152

chrisrosset:
> Most can probably be defaulted to null, when read though. This does seem to be common for the POST/GET variables, but not so consistent for other variables, for example like seen in #1152

Wouldn't this one also work `null` as the default?

reedy:
> > Most can probably be defaulted to null, when read though. This does seem to be common for the POST/GET variables, but not so consistent for other variables, for example like seen in #1152
> 
> Wouldn't this one also work `null` as the default?

[...]

Issue #1154, opened 2022-12-23
PHP Notice: Undefined index: id in flights.php on line 25


reedy:

reedy:
```
  $apid = $_POST["id"];
  if(! $apid) {
    $apid = $_GET["id"];
  }
```

line 25 is the `$_GET`... What should it default to?

PR #1207, opened 2023-06-19
Addressing many PHP 8 warnings

chrisrosset:
- Added the null coalescing operator (`??`) in tons of places to avoid warnings (notices before version 8) about undefined array keys. The most common use cases are reading optional request parameters or handling multiple different request types in the same place.
- `auto_detect_line_endings` is [deprecated in PHP 8](https://php.watch/versions/8.1/auto_detect_line_endings-ini-deprecated) and redundant.
- MySQL returns `null` for SUM on zero rows. Used `COALESCE` to get 0 instead for use in calculations. (scenario: User settings page for a user without any flights - e.g. a new user)
- Used `intdiv` to avoid "PHP deprecated: Implicit conversion from float ... to int loses precision" when calculation the total flight duration in days/hours/minutes for the banner.
- The `seattype` field is not defined while processing CSV imports. I set the value to `""` to avoid an undefined variable warning when the variable is used to present the user with the data read out of the file. Side note: I don't believe this field is actually used.
- Added `"" => ""` mappings in `import.php`'s' class/reason/pos lookup tables to avoid undefined array key warnings for cases.

### DIFF
--- a/badge/banner.php
+++ b/badge/banner.php
@@ -66,9 +66,13 @@ if ($row["public"] == "N") {
 $uid = $row["uid"];
 $units = $row["units"];
 
-$sth = $dbh->prepare(
-    "SELECT COUNT(*) AS count, SUM(distance) AS distance, SUM(TIME_TO_SEC(duration))/60 AS duration FROM flights WHERE uid=?"
-);
+$sth = $dbh->prepare(<<<SQL
+SELECT COUNT(*) AS count,
+       COALESCE(SUM(distance), 0) AS distance,
+       COALESCE(SUM(TIME_TO_SEC(duration))/60, 0) AS duration
+FROM flights
+WHERE uid=?
+SQL);
 $result = $sth->execute([$uid]);
 if ($result && $row = $sth->fetch()) {
     $distance = $row["distance"];
@@ -83,7 +87,7 @@ if ($result && $row = $sth->fetch()) {
     $duration = sprintf(
         "%d days, %d:%02d hours",
         $row["duration"] / 1440,
-        ($row["duration"] / 60) % 24,
+        intdiv($row["duration"], 60) % 24,
         $row["duration"] % 60
     );
 } else {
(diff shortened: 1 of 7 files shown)


chrisrosset:
Fixes #1154.
Fixes #1155.

php/stats.php
```diff
@@ -5,19 +5,13 @@
...
 include 'filter.php';
 
-$uid = $_SESSION["uid"];
 $public = "O"; // by default...
-if (!$uid || empty($uid)) {
-    // If not logged in, default to demo mode
-    $uid = 1;
-}
 
-// This applies only when viewing another users flights
```
reedy, line 15: Guess this comment wasn't removed on purpose (still exists in the top10.php code below)

badge/banner.php
```diff
@@ -83,7 +87,7 @@ function renderImage($textLines, $user = false) {
     $duration = sprintf(
         "%d days, %d:%02d hours",
         $row["duration"] / 1440,
-        ($row["duration"] / 60) % 24,
+        intdiv($row["duration"], 60) % 24,
```
reedy, line 90: This looks a bit weird to be (I'd probably be more tempted to use `(int)` casting), but very much a stylistic choice. So doesn't really matter.

I know some IDE's complain about some calls like this for performance reasons, but this doesn't seem to be one of them.

php/stats.php
reedy: ^

reedy (changes requested):

php/top10.php
```diff
@@ -29,10 +27,8 @@
         $mode = "COUNT(fid)";
         break;
 }
-$limit = $_POST["limit"];
-if (!$limit) {
-    $limit = "10";
-}
+
+$limit = $_POST["limit"] ?? "10";
```
reedy, line 31: These changes upwards conflicted with 282fe65d9481d917f40b95c94cf1fc27c205a409 which did some similar changes

reedy:
Rebased and amended; pushed as 52e054d87d63ba0d43153ff61ef8ed59b8fb0d3a

Issue #1197, opened 2023-04-23
Import issues


reedy:
On my dev install... I deleted all my flights (using the UI) and then imported a newer backup from openflights.org

It seems to have resulted in some weird plane type attributions, even though the exported CSV looks good

![Screenshot 2023-04-23 at 18 49 28](https://user-images.githubusercontent.com/67615/233856236-3553fea6-fc38-4493-8abf-8fed161e59e2.png)

Should this be showing IDs or types?

![Screenshot 2023-04-23 at 18 50 44](https://user-images.githubusercontent.com/67615/233856332-cd97fcdf-7353-4c22-a08b-3ddea4dcd2e5.png)

reedy:
I doubt the fix to https://github.com/jpatokal/openflights/issues/1116 is to blame....

reedy:
But I'm guessing something to do with differing plane ids locally..?

```
mysql> select * from planes;
+-------------------------+------+-------+------+--------+
| name                    | abbr | speed | plid | public |
+-------------------------+------+-------+------+--------+
| Airbus A320             | NULL |  NULL |    1 | NULL   |
| Boeing 777-300ER        | NULL |  NULL |    2 | NULL   |
| Boeing 737-MAX9         | NULL |  NULL |    3 | NULL   |
| Boeing 737-800          | NULL |  NULL |    4 | NULL   |
| Embraer ERJ175          | NULL |  NULL |    5 | NULL   |
| Airbus A319             | NULL |  NULL |    6 | NULL   |
| Airbus A321NEO          | NULL |  NULL |    7 | NULL   |
| Embraer E175            | NULL |  NULL |    8 | NULL   |
| Airbus A321             | NULL |  NULL |    9 | NULL   |
| Airbus A350-900         | NULL |  NULL |   10 | NULL   |
| Airbus A330-300         | NULL |  NULL |   11 | NULL   |
| Embraer E190            | NULL |  NULL |   12 | NULL   |
| Airbus A320NEO          | NULL |  NULL |   13 | NULL   |
| Embraer 190             | NULL |  NULL |   14 | NULL   |
| Boeing 777-200          | NULL |  NULL |   15 | NULL   |
| Boeing 787-8            | NULL |  NULL |   16 | NULL   |
| BAE Jetstream J41       | NULL |  NULL |   17 | NULL   |
| Airbus A380             | NULL |  NULL |   18 | NULL   |
| Embraer E190SR          | NULL |  NULL |   19 | NULL   |
[...]

Issue #1137, opened 2022-12-05
load-data.sql: ERROR 1054 (42S22) at line 30: Unknown column 'iso_code' in 'field list'


reedy:
Following on from https://github.com/jpatokal/openflights/issues/1103

```
Importing airlines...
Importing airports...
Importing routes...
Importing countries...
ERROR 1054 (42S22) at line 30: Unknown column 'iso_code' in 'field list'
```

reedy:
https://github.com/jpatokal/openflights/blame/master/sql/load-data.sql#L35

`(name, iso_code, dafif_code);`

vs

https://github.com/jpatokal/openflights/blob/master/sql/create.sql#L97-L105

```
DROP TABLE IF EXISTS `countries`;
CREATE TABLE `countries` (
  `junk` text,
  `code` varchar(2) NOT NULL,
  `name` text,
  `oa_code` varchar(2) default NULL,
  `dst` char(1) default NULL,
  PRIMARY KEY  (`code`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8;
```

I guess `iso_code, dafif_code` should be `code, oa_code` or `oa_code, code` ?

reedy:
Or maybe it's right?

And after `create-tables.sql`, `airlines-upgrade.sql` and `airports-upgrade.sql` need running? In which order? Which isn't documented, but would be much more helpful if `create-tables.sql` was actually canonical...

reedy:
Can we get a dump of the current schema structure from the live site?

jpatokal:
Here's the live DB:

```
CREATE TABLE `countries` (
  `name` text,
  `iso_code` varchar(2) DEFAULT NULL,
  `dafif_code` varchar(2) NOT NULL DEFAULT '',
  PRIMARY KEY (`dafif_code`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8

CREATE TABLE `airlines` (
  `name` varchar(80) DEFAULT NULL,
  `iata` varchar(2) DEFAULT NULL,
  `icao` varchar(3) DEFAULT NULL,
  `callsign` text,
  `country` text,
  `country_code` varchar(2) DEFAULT NULL,
  `alid` int(11) NOT NULL AUTO_INCREMENT,
  `uid` int(11) DEFAULT NULL,
  `alias` text,
  `mode` char(1) DEFAULT 'F',
  `active` varchar(1) DEFAULT 'N',
  `source` text,
[...]

PR #1211, opened 2023-06-23
SQL updates for the `planes` table

### LINKED ISSUE #1210: `load-data.sql` does not load seed the `planes` table
The data is present in `data/planes.dat` but is not loaded.

chrisrosset:
This updates `create.sql` with the schema changes that were brought in through `clean-duplicate-planes.sql` in https://github.com/jpatokal/openflights/commit/36b3ee9d5c087e38f9011c9c338a37f7ddb06ee2.

Additionally, this updates `load-data.sql` to seed the DB with the data present in `data/planes.dat`.

Resolves #1210.

### DIFF
--- a/sql/create.sql
+++ b/sql/create.sql
@@ -1,5 +1,7 @@
+DROP DATABASE IF EXISTS flightdb2;
 CREATE DATABASE flightdb2;
 
+DROP USER IF EXISTS openflights@localhost;
 CREATE USER openflights@localhost;
 GRANT ALL PRIVILEGES ON flightdb2.* TO openflights@localhost;
 
@@ -155,12 +157,16 @@ CREATE TABLE `locales` (
 
 DROP TABLE IF EXISTS `planes`;
 CREATE TABLE `planes` (
-  `name` text,
+  `name` varchar(80),
   `abbr` text,
   `speed` double default NULL,
   `plid` int(11) NOT NULL auto_increment,
   `public` char(1) default NULL,
-  PRIMARY KEY  (`plid`)
+  `iata` text default NULL,
+  `icao` text default NULL,
+  `frequency` int default 0,
+  PRIMARY KEY  (`plid`),
+  UNIQUE KEY `name` (`name`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;
 
 DROP TABLE IF EXISTS `routes`;
--- a/sql/create.sql
+++ b/sql/create.sql
@@ -161,10 +161,10 @@ CREATE TABLE `planes` (
   `abbr` text,
   `speed` double default NULL,
   `plid` int(11) NOT NULL auto_increment,
-  `public` char(1) default NULL,
+  `public` char(1) default 'N',
   `iata` text default NULL,
   `icao` text default NULL,
-  `frequency` int default 0,
+  `frequency` int(11) default 0,
   PRIMARY KEY  (`plid`),
   UNIQUE KEY `name` (`name`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;
--- a/sql/create.sql
+++ b/sql/create.sql
@@ -1,7 +1,5 @@
-DROP DATABASE IF EXISTS flightdb2;
 CREATE DATABASE flightdb2;
 
-DROP USER IF EXISTS openflights@localhost;
 CREATE USER openflights@localhost;
 GRANT ALL PRIVILEGES ON flightdb2.* TO openflights@localhost;
 
--- a/sql/clean-duplicate-planes.sql
+++ b/sql/clean-duplicate-planes.sql
@@ -7,7 +7,7 @@ select name, min(plid), max(plid), count(*) as dupes from planes GROUP BY name h
 
 -- Find exact dupes
 drop table if exists tmp_planes_duplicates;
-create table tmp_planes_duplicates (plid int, duplicate_of int)
+create table tmp_planes_duplicates (plid int, duplicate_of int);
 insert into tmp_planes_duplicates
     select p1.plid, min(p2.plid) as duplicate_of
     from planes p1, planes p2
--- a/sql/load-data.sql
+++ b/sql/load-data.sql
@@ -34,6 +34,14 @@ OPTIONALLY ENCLOSED BY '"'
 LINES TERMINATED BY '\n'
 (name, iso_code, dafif_code);
 
+\! echo Importing planes...
+LOAD DATA LOCAL INFILE 'data/planes.dat'
+REPLACE INTO TABLE planes
+FIELDS TERMINATED BY ','
+OPTIONALLY ENCLOSED BY '"'
+LINES TERMINATED BY '\n'
+(name, iata, icao);
+
 \! echo Importing locales...
 
 LOAD DATA LOCAL INFILE 'locale/locales.dat'


chrisrosset:
@jpatokal If you ever get a moment, perhaps you could share a schema dump (I'm not a MySQL expert but I think `mysqldump --no-data flightdb2` would do the trick) of all the tables in the database?

I remember @reedy updated `create.sql` for `countries`, `airports` and `airlines` based on the schema you shared in https://github.com/jpatokal/openflights/issues/1137#issuecomment-1339376478. I would gladly update all the other ones.

chrisrosset:
I just noticed that #1148 did not include the `frequency` column from https://github.com/jpatokal/openflights/issues/1137#issuecomment-1339376478. @reedy, was this on purpose?

reedy:
> I just noticed that #1148 did not include the `frequency` column from [#1137 (comment)](https://github.com/jpatokal/openflights/issues/1137#issuecomment-1339376478). @reedy, was this on purpose?

Not particularly. Probably mostly because it wasn't throwing an error...

I'd be curious if it's even used, as it seems to be (maybe) updated uses a script that pulls it in from VRS - https://github.com/jpatokal/openflights/blob/2f86a25d09c0721a2b37aa464bd7c474244fdeab/tools/import-vrs.py

reedy:
Though as per https://github.com/jpatokal/openflights/pull/1148 "create.sql: Integrate *-upgrade.sql", not "update to match live schema"...

https://github.com/jpatokal/openflights/blob/4ff8baf/sql/airlines-upgrade.sql doesn't have a `frequency` column.

chrisrosset:
> Though as per #1148 "create.sql: Integrate *-upgrade.sql", not "update to match live schema"...

Fair enough.

The `clean-duplicate-*.sql` scripts bring in more columns (`iata`, `icao`, and `frequency`):

https://github.com/jpatokal/openflights/blob/master/sql/clean-duplicate-airlines.sql#L28
https://github.com/jpatokal/openflights/blob/master/sql/clean-duplicate-planes.sql#L33

> I'd be curious if it's even used, as it seems to be (maybe) updated uses a script that pulls it in from VRS - https://github.com/jpatokal/openflights/blob/2f86a25d09c0721a2b37aa464bd7c474244fdeab/tools/import-vrs.py

I don't know if they're kept up to date but they are read by the code and produce SQL errors when not present:

* https://github.com/jpatokal/openflights/blob/ea99f97df79eb8d9f279c4de9d95a650cea2a95c/php/import.php#L169
* https://github.com/jpatokal/openflights/blob/ea99f97df79eb8d9f279c4de9d95a650cea2a95c/php/autocomplete.php#L191

reedy:
import wasn't broken (at least, not with an SQL error related to this) on my dev wiki, even without (presumably) having the `frequency` column (I have just checked my live schema, and it's indeed not there, so not apparently added by a random patch I applied).

chrisrosset:
> import wasn't broken (at least, not with an SQL error related to this) on my dev wiki, even without (presumably) having the `frequency` column (I have just checked my live schema, and it's indeed not there, so not apparently added by a random patch I applied).

This code only runs when the import is trying to match names based on strings. I don't know if you saw https://github.com/jpatokal/openflights/issues/1197#issuecomment-1605185457 but I'm guessing you were using an import file with ID columns populated by the export/backup.

Here's an example SQL warning:

`PHP Warning:  PDOStatement::execute(): SQLSTATE[42S22]: Column not found: 1054 Unknown column 'frequency' in 'order clause' in /var/www/openflights/php/import.php on line 172`

reedy:
> > import wasn't broken (at least, not with an SQL error related to this) on my dev wiki, even without (presumably) having the `frequency` column (I have just checked my live schema, and it's indeed not there, so not apparently added by a random patch I applied).
> 
> This code only runs when the import is trying to match names based on strings. I don't know if you saw [#1197 (comment)](https://github.com/jpatokal/openflights/issues/1197#issuecomment-1605185457) but I'm guessing you were using an import file with ID columns populated by the export/backup.

Anything running the "newer" code would indeed have the column populated.

I don't know from what version of openflights I would've originally exported the backup from.

But TLDR is that I was using my "live" data from openflights.org, into my dev install.

Makes it easy to compare changes...

I probably need to wipe the entire database anyway

jpatokal:
@chrisrosset The frequency column is not used by the live website, IIRC it was there to try to sort out duplicates.  Here's what the table looks like on the live system:

```
mysql> describe planes;
+-----------+-------------+------+-----+---------+----------------+
| Field     | Type        | Null | Key | Default | Extra          |
+-----------+-------------+------+-----+---------+----------------+
| name      | varchar(80) | YES  | UNI | NULL    |                |
| abbr      | text        | YES  |     | NULL    |                |
| speed     | double      | YES  |     | NULL    |                |
| plid      | int(11)     | NO   | PRI | NULL    | auto_increment |
| public    | char(1)     | YES  |     | N       |                |
| iata      | text        | YES  |     | NULL    |                |
| icao      | text        | YES  |     | NULL    |                |
| frequency | int(11)     | YES  |     | 0       |                |
+-----------+-------------+------+-----+---------+----------------+
8 rows in set (0.00 sec)
```

sql/create.sql
```diff
@@ -155,12 +157,16 @@ CREATE TABLE `locales` (
...
-  `name` text,
+  `name` varchar(80),
   `abbr` text,
   `speed` double default NULL,
   `plid` int(11) NOT NULL auto_increment,
   `public` char(1) default NULL,
-  PRIMARY KEY  (`plid`)
+  `iata` text default NULL,
+  `icao` text default NULL,
+  `frequency` int default 0,
```
reedy: looks like this should be `int(11)` to match the live schema

sql/create.sql
```diff
@@ -155,12 +157,16 @@ CREATE TABLE `locales` (
 
 DROP TABLE IF EXISTS `planes`;
 CREATE TABLE `planes` (
-  `name` text,
+  `name` varchar(80),
   `abbr` text,
   `speed` double default NULL,
   `plid` int(11) NOT NULL auto_increment,
   `public` char(1) default NULL,
```
reedy: `default 'n'`?

chrisrosset:
@jpatokal, many thanks for posting the schema.

> @chrisrosset The frequency column is not used by the live website, IIRC it was there to try to sort out duplicates. Here's what the table looks like on the live system:

The columns are used! I've even already posted the links to where they're used above. :) I'll list them in more detail now:

1. [Airline frequency is used during import to sort airlines when matching by name.](https://github.com/jpatokal/openflights/blob/ea99f97df79eb8d9f279c4de9d95a650cea2a95c/php/import.php#L169)
   ```sql
   $sql = "select name,alias,alid from airlines
   where ((name like ? or alias like ?) and (iata != '')) or (name = ?) order by frequency
   ```
2. [Plane frequency is used for autocomplete.](https://github.com/jpatokal/openflights/blob/ea99f97df79eb8d9f279c4de9d95a650cea2a95c/php/autocomplete.php#L191) The code and the comment both reference frequency.
   ```php
    // Autocompletion for plane types
    // First match against major types with IATA codes, then pad to max 6 by matching against frequency of use
   $query = $_POST['plane'];
   $name = "%$query%";
   $query = "(name LIKE :name OR iata LIKE :name) ";
   $sql = "(SELECT name,plid FROM planes WHERE " . $query . " AND iata IS NOT NULL ORDER BY name LIMIT 6) UNION " .
   "(SELECT name,plid FROM planes WHERE " . $query . " AND iata IS NULL ORDER BY frequency DESC LIMIT 6) LIMIT 6";
   ```

Based on them being used, I think we should include them in `create.sql`. I'll update this PR to match the live schemas posted here and in the previous PR.

It'd be nice to get a `mysqldump --no-data` to confirm we haven't missed anything. :)

reedy:
I'd be curious if any indexes are out of line too

chrisrosset:
@reedy, anything else here? If we do get a full schema dump, I can do another pass in a separate PR. I'd like to get this merged because the Docker PR (#1206) has this changed merged in already.

sql/create.sql
```diff
@@ -1,5 +1,7 @@
+DROP DATABASE IF EXISTS flightdb2;
 CREATE DATABASE flightdb2;
 
+DROP USER IF EXISTS openflights@localhost;
```
reedy: Not that it set one before... passwordless is fun :).

Obviously doesn't matter much for a dev type install.

Probably should be punted somewhere else for documentation updates though. Doesn't necessarily need fixing in this commit

sql/create.sql
```diff
@@ -1,5 +1,7 @@
+DROP DATABASE IF EXISTS flightdb2;
```
reedy: Destructive!

Arguably all the individual `DROP TABLE IF EXISTS` are now redundant... Or your mariadb/mysql is buggy AF if it manages to drop tables at those points

reedy (changes requested):
There's a little bit of change in the "functionality" of create.sql now... Not that I'm against it.

But I think https://github.com/jpatokal/openflights/blob/52e054d87d63ba0d43153ff61ef8ed59b8fb0d3a/sql/README#L4 and maybe https://github.com/jpatokal/openflights/blob/52e054d87d63ba0d43153ff61ef8ed59b8fb0d3a/widget/index.html#L48-L50 should reference that it is destructive (I know it already was, as it would drop each table)

sql/create.sql
chrisrosset: They were redundant before this change too since `create database` would error out if one already existed. But if it didn't, then it'd be destructive since it would drop all the tables.

Perhaps the other changes shouldn't have been merged back into `create.sql` and the scripts should have been run as database migrations...

sql/create.sql
chrisrosset: I'll remove this and the drop db from the PR.

chrisrosset:
> and maybe
> 
> https://github.com/jpatokal/openflights/blob/52e054d87d63ba0d43153ff61ef8ed59b8fb0d3a/widget/index.html#L48-L50
> should reference that it is destructive (I know it already was, as it would drop each table)

I think the widget stuff is all dead code. I opened #1224 to ask @jpatokal about removing it.

PR #1213, opened 2023-06-24
Improved Banner Thousands Separator Logic

### LINKED ISSUE #153: [Bug] Thousands separator problem [sf#153]
**Reported by *anonymous on 2013-04-03 11:17 UTC**
When going over 1.000.000 km/miles, the thousands separator is no longer inserted (e.g., it looks like 1000,000). See http://openflights.org/banner/Imperator1972.png

chrisrosset:
Use `number_format` to generalize inserting the thousands separators. This fixes formatting for distances over 1 million.

Tested numbers up to 1 billion.

| Old format   | New format     |
|--------------|----------------|
|            1 |              1 |
|        1,000 |          1,000 |
|     1000,000 |      1,000,000 |
|  1000000,000 |  1,000,000,000 |

Fixes #153.

Side notes:

1. We specify "," because the banner is in English. We could load the user's locale and use that but that would only make sense if we internationalized the whole banner.
2. The web UI doesn't have any thousands separator at all. Easy enough to add but ought to be done in a separate PR.

### DIFF
--- a/badge/banner.php
+++ b/badge/banner.php
@@ -79,7 +79,7 @@ if ($result && $row = $sth->fetch()) {
         $units = "miles";
     }
     $flights = sprintf("%s flights", $row["count"]);
-    $miles = sprintf("%d,%03d %s", $distance / 1000, $distance % 1000, $units);
+    $miles = sprintf("%s %s", number_format($distance, 0, ".", ","), $units);
     $duration = sprintf(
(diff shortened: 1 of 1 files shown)


reedy:
i18n/l10n-ify-ing the banner would be nice, but out of scope... And this would make it easier anyway!

PR #1214, opened 2023-06-24
Fix Artistic Layer Functionality

chrisrosset:
The artistic layer (upper right hand side globe button -> artistic) seems to be broken on https://openflights.org today. The map does not load any tiles.

In the debug console, I'm seeing all tile requests return 404s (they first get 302'd to https://tiles.stadiamaps.com).

[The docs for the Watercolor tiles](http://maps.stamen.com/#watercolor) we're using for this specify `.jpg` (instead of `.png`) as the extension in the URL template.

I checked Internet Archive snapshots and the `jpg` extension has been there [since 2015](https://archive.is/hFtwi). My theory is that they used to serve `png` as well but perhaps they stopped when they set up the redirects to stadiamaps.com.

### DIFF
--- a/openflights.js
+++ b/openflights.js
@@ -137,7 +137,7 @@ var poliLayer = new OpenLayers.Layer.XYZ(
 var artLayer = new OpenLayers.Layer.XYZ(
     "Artistic",
     [
-        "https://stamen-tiles.a.ssl.fastly.net/watercolor/${z}/${x}/${y}.png"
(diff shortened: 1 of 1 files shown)


reedy:
This is curious... I'm pretty sure this was working when I was playing with the layers a couple of months ago, so must be a recent breakage...

reedy:
Confirmed fixed 👍

PR #1218, opened 2023-06-28
nginx: Fix access to profiles with . in name

### LINKED ISSUE #1167: Can't access user profile - problem with period in name?
My OpenFlights user name is adam.smith. I'm pretty sure it used to be possible for people to access my profile using the standard URL of https://openflights.org/user/adam.smith. But it has recently been throwing up a 404 Not Found error. I'm wondering whether some recent change has create a problem with the period in the name?

reedy:
Fixes https://github.com/jpatokal/openflights/issues/1167

### DIFF
--- a/nginx/openflights
+++ b/nginx/openflights
@@ -24,27 +24,21 @@ server {
 	rewrite "^/banner/(.+).png$" /badge/banner.php?user=$1;
     }
 
-    if ($uri !~ \.) {
-	rewrite "^/airline/(.+)$" /index.php?query=$1;
-	rewrite "^/airport/(.+)$" /index.php?query=$1;
-	rewrite "^/blog/(.+)$" https://blog.openflights.org/$1;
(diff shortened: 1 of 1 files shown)


nginx/openflights
```diff
@@ -24,27 +24,21 @@ server {
 	rewrite "^/banner/(.+).png$" /badge/banner.php?user=$1;
     }
 
-    if ($uri !~ \.) {
```
reedy, line 27: I'm not quite sure what this is supposed to be preventing...

It does a negated regex match looking for a literal `.`... But when usernames can have `.` in them..

Maybe airline and airport shouldn't be in it, but what does that give?

reedy: @jpatokal Can you remember what you added this for?

jpatokal: My best guess would be that it's trying to avoid rewriting file paths that already have .php/html, but the regexes should be specific enough to avoid that.

Issue #1216, opened 2023-06-28
Display version (gith hash) of OpenFlights on website somewhere


reedy:
Not sure if we want to display it on the About page, or as some sort of footer somewhere...

But it'd be useful to be able to easily tell what version of the site is actually deployed

PR #1219, opened 2023-06-29
Include git commit hash in HTML metadata

chrisrosset:
An implementation for https://github.com/jpatokal/openflights/issues/1216.

This assumes the deployment point is a git clone.

### DIFF
--- a/about.php
+++ b/about.php
@@ -1,7 +1,12 @@
+<?php
+require_once "php/git.php";
+?>
+
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
 <html>
 <head>
     <title>OpenFlights: About</title>
+    <meta name="version" content="<?php echo Git::getCurrentCommitID() ?? "unavailable"; ?>">
     <link rel="stylesheet" href="/css/style_reset.css" type="text/css">
     <link rel="stylesheet" href="/openflights.css" type="text/css">
     <link rel="icon" type="image/png" href="/img/icon_favicon.png"/>
--- a/php/git.php
+++ b/php/git.php
@@ -1,0 +1,19 @@
+<?php
+
+class Git {
+
+    /**
+     * Retrieve the commit ID of the current HEAD.
+     * @return string|null commit ID or null on failure
+     */
+    public static function getCurrentCommitID() {
+        $head = file_get_contents(".git/HEAD");
+
+        $matches = [];
+        if (preg_match("/ref: (.*)/", $head, $matches)) {
+            $head = file_get_contents(".git/{$matches[1]}");
+        }
+
+        return $head ? $head : null;
+    }
+}
--- a/data.html
+++ b/data.html
@@ -147,7 +147,7 @@
             </tr>
         </table>
 
-        <p>The GitHub copy is only a sporadically updated static snapshot of the live OpenFlights database (see <a href="https://github.com/jpatokal/openflights/commits/master/data/airports.dat">revision log</a>). If you would like an up-to-the-minute copy, or you would like your data filtered by any information available to us (eg. number of routes at the airport), do not hesitate to <a href="/about.html">contact us</a>.<p>
+        <p>The GitHub copy is only a sporadically updated static snapshot of the live OpenFlights database (see <a href="https://github.com/jpatokal/openflights/commits/master/data/airports.dat">revision log</a>). If you would like an up-to-the-minute copy, or you would like your data filtered by any information available to us (eg. number of routes at the airport), do not hesitate to <a href="/about">contact us</a>.<p>
 
         <p>If you'd like an even more thorough database, with extensive coverage of airstrips, heliports and other places of less interest for commercial airline frequent flyers, do check out <a href="https://ourairports.com">OurAirports</a>, whose public domain database covers over 40,000 places to fly from.<p>
 
@@ -243,7 +243,7 @@
             </tr>
         </table>
 
-        <p>The GitHub copy is only a sporadically updated static snapshot of the live OpenFlights database (see <a href="https://github.com/jpatokal/openflights/commits/master/data/airlines.dat">revision log</a>). If you would like an up-to-the-minute copy, or you would like your data filtered by any information available to us (eg. number of flights by airline), do not hesitate to <a href="/about.html">contact us</a>.<p>
+        <p>The GitHub copy is only a sporadically updated static snapshot of the live OpenFlights database (see <a href="https://github.com/jpatokal/openflights/commits/master/data/airlines.dat">revision log</a>). If you would like an up-to-the-minute copy, or you would like your data filtered by any information available to us (eg. number of flights by airline), do not hesitate to <a href="/about">contact us</a>.<p>
 
         <a name="route"></a>
         <h2>Route database</h2>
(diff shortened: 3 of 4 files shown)


about.php
```diff
@@ -2,6 +2,7 @@
 <html>
 <head>
 	<title>OpenFlights: About</title>
+    <meta name="version" content="<?php echo(`git rev-parse HEAD`); ?>">
```
chrisrosset: Important note: `git` will refuse to run if the repo is not owned by $USER.

reedy: Can we indent it consistently please? ;)

reedy (changes requested):

about.php
chrisrosset: Taaaaaabs! \*shakes fist\*

More importantly, I'm afraid this won't work with the setup described in [INSTALL](https://github.com/jpatokal/openflights/blob/master/INSTALL). The instructions do the checkout as `openflights` but the server will be running as `www-data` so `git` will refuse to run. In my local testing the files were owned by `www-data`. Womp Womp.

Here are some options:

1. We could silence the ownership warning by running this as part of the install:
   ```sh
   sudo -u www-data git config --global --add safe.directory /var/www/openflights
   ```
   This would also have to be run on production. I think this makes it less likely to happen. :)

2. A post-checkout hook dumping the hash to a file would also work but, again, this also requires enabling the hook on the clone.

3. My previous attempt (`echo(file_get_contents(".git/refs/heads/master")`) doesn't need this but assumes `master` is the checked out branch.

4. A more general version would run `cat .git/$(cat .git/HEAD | cut -d' ' -f2)` which now works with different branches (but not a detached HEAD).

@reedy, thoughts or preferences? Do you know of a better way?

reedy: We've got a class in MediaWiki for dealing with these various edge cases, unfortunately it's not very portable - https://github.com/wikimedia/mediawiki/blob/master/includes/GitInfo.php

I suspect the answer here may be "don't just try and do it as a one liner".

But also, most people aren't probably going to be running their own version of openflights (unlikes someone might with MediaWiki), so potentially something that works trivially enough on a dev install, but will also work on the actual deployment (I don't have access, so I don't really know how it looks).

php/git.php
```diff
@@ -0,0 +1,20 @@
...
+
+        if (preg_match("/ref: (.*)/", $head, $matches)) {
+            $head = file_get_contents(".git/{$matches[1]}");
+        }
+
+        return $head ? $head : null;
+    }
+}
+
+?>
```
reedy: Don't need to have `?>` at the end of files; it's more likely to cause problems.

php/git.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+class Git {
+
+    /**
+     * Retrieve the commit ID of the current HEAD.
+     * @return string|null commit ID or null on failure
+     */
+    public static function getCurrentCommitID() {
+        $head = file_get_contents(".git/HEAD");
```
reedy, line 10: `file_get_contents` can return false, but `preg_match` doesn't seem to care (at least not atm).

php/git.php
```diff
@@ -0,0 +1,20 @@
...
+
+        if (preg_match("/ref: (.*)/", $head, $matches)) {
```
reedy, line 13: Not a bad thing to explicitly define `$matches` before the function

about.php
```diff
@@ -1,7 +1,12 @@
+<?php
+require_once("php/git.php")
```
reedy: While it can be used as one, it's not actually a function - https://www.php.net/manual/en/function.require-once.php

And should have a trailing `;` too.

`require_once "php/git.php";`

reedy (changes requested):

about.php
chrisrosset: I really appreciate these comments - I'm new to PHP. Will fix.

php/git.php
chrisrosset: done

php/git.php
chrisrosset: FWIW, that MediaWiki class does the same thing

https://github.com/wikimedia/mediawiki/blob/master/includes/GitInfo.php#L215

reedy: Yeah. My comment wasn't necessarily anything to fix; more me reviewing the code and leaving notes for myself etc too

php/git.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+class Git {
```
reedy, line 3: Wonder if we should just namespace this from day 1?

about.php
```diff
@@ -1,7 +1,12 @@
+<?php
+require_once "php/git.php";
+?>
+
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
 <html>
 <head>
 	<title>OpenFlights: About</title>
+	<meta name="version" content="<?php echo(Git::getCurrentCommitID() ?? ""); ?>">
```
reedy: My only outstanding question here is whether having a version of an empty string would look odd, but in practice, I imagine it's unlikely to happen...

chrisrosset: `unavailable`?

php/git.php
chrisrosset: You mean instead of using a class, right? I'm fine either way - this isn't really a class anyway.

reedy: I was meaning as well as.

It probably is a little overkill having a class for one function, sure, but at some point we can rename it to something like Utils when more things are being created/refactored.

Not like we really need to worry about "back compat" etc.

about.php
reedy: Doesn't look like this one got fixed... Or got lost :)

reedy (changes requested):
If we're renaming it from `about.html` to `about.php`, I suspect there's a few links that need/want updating too

chrisrosset:
The reformatting is killing me. :)

chrisrosset:
> If we're renaming it from `about.html` to `about.php`, I suspect there's a few links that need/want updating too

FWIW, it works regardless because of our nginx setup.

I removed the extensions. The majority of about links seems to already be extensionless and relies on the nginx rewrite rules.

about.php
```diff
@@ -1,7 +1,12 @@
+<?php
+require_once "php/git.php";
+?>
+
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
 <html>
 <head>
     <title>OpenFlights: About</title>
+    <meta name="version" content="<?php echo(Git::getCurrentCommitID() ?? "unavailable"); ?>">
```
reedy: echo isn't a function either (like `require_once`) - https://www.php.net/manual/en/function.echo.php#refsect1-function.echo-notes

chrisrosset:
While we're here, which one is preferred: echo or print?

about.php
chrisrosset: > echo isn't a function either (like `require_once`) - https://www.php.net/manual/en/function.echo.php#refsect1-function.echo-notes

Would you prefer the short form syntax here?

```
<?=(Git::getCurrentCommitID() ?? "unavailable")?>
```

I kind of like the parenthesis in this form because it delimits the expression more clearly.

reedy:
> While we're here, which one is preferred: echo or print?

Most openflights code uses `echo`, but `printf` when formatted needed.

Keeping with echo would keep things consistent.

reedy:
I really don't know why GitHub thinks https://github.com/jpatokal/openflights/commit/eb830a82787939a1ca999b0b406da88f82dc8280#diff-2212b84a2a6ffe239872c76d266680d86f2438391034b2eb19604f17104f6968 conflicts...

Git locally just deals with it fine

reedy:
a70afa112267e4f0f21bd3ef516f690958facf93

Issue #876, opened 2018-07-31
REST API / Redesign


blumk:
I took a quick look at the source code and here's my proposal for untangling the code:
- Back-end with a proper REST API
- Front-end that uses the REST API
And both back-end/front-end are completely independent of each other.

Since rewriting the whole application is one big effort, I propose feature slicing (one use case at a time):
Example:
- Back-end: GET flights for one user (public profile)
- Front-end: Sample site that consumes this data and displays it

@jpatokal If you are interested in redesigning this software, I'm more than willing to help you out with the front-end part (user experience, coding, progressive web app).

a2hill:
I was looking around the issues for a roadmap or any kind of list of help wanted/features and this seems like a good a place to start as any. I too am willing to help out with this as I really enjoy the product and would love to have an API! @blumk @jpatokal Any interest in tackling this or at least putting together some kind of documentation for willing contributors to help out with?

jpatokal:
These are very good questions and I've put together a very high-level roadmap to address them. comments are very welcome.

https://docs.google.com/document/d/1yyNB-yDLVewyNpJ78ZvC3w6K3MB1NUvBFdNNFuET-9c/edit#

PR #1222, opened 2023-06-30
Return JSON-encoded results from `top10.php`

chrisrosset:
`top10.php` now returns JSON for both the data and errors. This change also includes the front-end side which consumes the new format.

A baby step towards https://github.com/jpatokal/openflights/issues/876.

### DIFF
--- a/openflights.js
+++ b/openflights.js
@@ -1698,9 +1698,18 @@ function googleChart(targetdiv, inputdata, labeldata) {
   chart.draw(data, GOOGLE_CHART_OPTIONS);
 }
 
-function showTop10(str) {
-  if(str.substring(0,5) == "Error") {
-    $("result").innerHTML = str.split(';')[1];
+function showTop10(responseText) {
+  let topData;
+  try {
+    topData = JSON.parse(responseText);
+  } catch (e) {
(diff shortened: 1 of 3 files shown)


openflights.js
```diff
@@ -1715,94 +1724,62 @@ function showTop10(str) {
...
-        var airportName = col.slice(0, tooManyColumns + 1).join(',');
-        desc = airportName.substring(0,20) + " (" + col[noOfColumns - 3] + ")";
-      }
-      else {
-        desc = col[0].substring(0,20) + " (" + col[1] + ")";
-      }
-      table += "<tr><td><a href='#' onclick='JavaScript:selectAirport(" + col[noOfColumns-1] + ");'>" + desc + "</a></td><td style='text-align: right; padding-left: 10px'>" + parseInt(col[noOfColumns-2]) + "</td>";
-    }
-    table += "</table>";
+  bigtable += "<td style='vertical-align: top; background-color: #ddd; padding: 0px 10px'>";
```
chrisrosset, line 1734: I kept the HTML generation the same (for now) but I actually have a separate branch converted to use `document.createElement`. Take a look at https://github.com/chrisrosset/openflights/compare/top10-json...chrisrosset:openflights:rewrite-showtop10?expand=1 if you're interested in what that looks like.

Issue #1192, opened 2023-04-07
Upgrade OpenLayers.js


reedy:
I'm under no illusion this will be easy/simple...

https://github.com/jpatokal/openflights/commit/99df5758b28b1b0c618eafc46d900a71fbacb5bd upgraded to 2.11 back in November 2011.

https://github.com/openlayers/openlayers/releases/tag/v7.3.0 was released in early March 2023
https://github.com/openlayers/openlayers/releases/tag/v7.4.0 was released in in early June 2023

May help with https://github.com/jpatokal/openflights/issues/916 maybe...

reedy:
https://github.com/openlayers/openlayers/compare/release-2.11...v2.13.1 is a lot of changes..

chrisrosset:
> I'm under no illusion this will be easy/simple...

You weren't kidding... :(

I've done some work for this on the side to find any major roadblocks. I have the basic map display working. Here's the branch if anyone's interested: https://github.com/chrisrosset/openflights/tree/openlayers7

## OpenLayers 2 (master)
![openlayers-old](https://github.com/jpatokal/openflights/assets/1103294/bc025f0c-ab68-4f5b-a485-634a5da55f16)


## OpenLayers 7 pre-alpha
![openlayers-new](https://github.com/jpatokal/openflights/assets/1103294/ad930d42-37a7-4973-8472-31b3ddf13bfd)

Notice the [dateline display issue](https://github.com/jpatokal/openflights/issues/2) is resolved. A lot functionality is missing still:

- cluster behavior is simplified for now
- popups are not implemented
- the code is super inefficient about how it deals with styles

## `prototype.js`-related breakage

I've noticed that the OpenLayers v7 attribution control is broken by `prototype.js`. I don't know why but that library sets up tons of stuff on core prototypes and is likely interfering with some standard JS functionality.

Unfortunately, we have a dependency chain which is not trivial to remove:

* `controls.js` (used for the autocomplete)
   * `scriptalicious.js`
      * `prototype.js`

reedy:
You've been busy!

At least we don't have jQuery to deal with too ;)

[scriptaculous](https://github.com/madrobby/scriptaculous) and [scriptaculous](https://github.com/madrobby/scriptaculous) don't seem overly active.. :/

Which means there's not really any updates in that chain that is going to really help

https://github.com/prototypejs/prototype/compare/1.7...1.7.3 which you've done already

[...]

PR #1223, opened 2023-07-01
UI Autocomplete Rework

chrisrosset:
> It would be interesting if we could just replace that all wholesale... https://github.com/jpatokal/openflights/issues/1192#issuecomment-1615614135

@reedy, ask and you shall receive! But now I need your review and help with testing.

@jpatokal, I know you're super busy but I think this is a neat upgrade and worth a review.

## What

This sizeable PR rips out the current autocomplete control with its two big legacy dependencies

1. `prototype.js`[^1] (no updates since 2015)
2. `scriptalicious.js` (no updates since 2010)

and replaces them with the much much smaller, modern, and actively maintained https://github.com/kraaden/autocomplete. The end result replicates the functionality almost exactly [^2]. I've chosen to use the stylesheets from `autocomplete` without any changes so there are minor changes in the look-and-feel.

Here's a quick video preview:
https://github.com/jpatokal/openflights/assets/1103294/cbbc895e-5508-4f8f-8c45-cdda68d8203d 

## Why

My main reason for this change is that the current autocomplete framework blocks multiple other improvements.

### JSON-encoded API responses

`autocomplete.php` currently returns an HTML fragment as the response which is then embedded into the DOM.

Request payload: `src_ap=new`

Response payload:

```html
<ul class='autocomplete'><li class='autocomplete' origin='src_ap' id='NEW:9188:-90.028297424316:30.042400360107:-6:A'>New Orleans-Lakefront Airp. (NEW), United States</li>
<li class='autocomplete' origin='src_ap' id='EWB:5737:-70.95690155029297:41.67610168457031:-5:A'>New Bedford Regional Airpo. (EWB), United States</li>
<li class='autocomplete' origin='src_ap' id='EWN:3730:-77.04290008539999:35.0730018616:-5:A'>New Bern-Coastal Carolina. (EWN), United States</li>
<li class='autocomplete' origin='src_ap' id='HVN:4006:-72.88680267:41.26369858:-5:A'>New Haven-Tweed New Haven. (HVN), United States</li>
<li class='autocomplete' origin='src_ap' id='MSY:3861:-90.25800323486328:29.99340057373047:-6:A'>New Orleans-Louis Armstron. (MSY), United States</li>
<li class='autocomplete' origin='src_ap' id='NBG:3768:-90.03500366:29.82530022:-6:A'>New Orleans NAS JRB/Alvin. (NBG), United States</li>
</ul>
```

This is super tight coupling between the UI and the server. This UI dependency has to be removed before we can upgrade the endpoint.

### OpenLayers 7 Upgrade

Whilst working on [upgrading the map to use OpenLayers 7](https://github.com/jpatokal/openflights/issues/1192) I discovered that `prototype.js` breaks some OL features (e.g. attribution controls). `prototype.js` does lots of low-level monkey patching all over the place which interferes with now standard functionality.

## How

I created a wrapper function around `autocomplete` to make the initialization resemble the old code.

I use `fetch` to send the request to `autocomplete.php` but then process the response as data instead of embedding it in the DOM. I parse the response as XML using `DOM.Parser` (this part would become `JSON.parse()` once the API starts returning JSON) and pass the data to `autocomplete.js` which deals with showing the list of items.

I used the vanilla CSS provided by the library.

[^1]: We still vendor and use `prototype.js` for the map widget. That's a whole other can of worms.
[^2]: This includes issues that exist in the current production version. I've managed to break the old code in lots of ways that don't really matter in normal usage.


js/autocomplete.js
reedy, line 1: Can we name it `autocomplete.min.js`? :)

js/compact.sh
```diff
@@ -1,5 +1,3 @@
 #!/bin/sh
 # Crunch the .orig files to save space and loading time
 jscompact -la <utilities.js.orig >utilities.js
-jscompact -la <scriptaculous.js.orig >scriptaculous.js
-
```
reedy: I've removed this already because the `.orig` doesn't exist

reedy (commented):
Minor annoyance with the whitespace changes too ;)

Sorry for the various conflicts

js/autocomplete.js
reedy: And can we include the version of the library used in one of the relevant commit summaries?

Makes chasing it in future (until we have better package management), and knowing when we may need/want to update.

https://www.npmjs.com/package/autocompleter shows 8.0.4 currently (but no tags in the github instance)

reedy: And you can include the minified and the non minified versions if you want...

We really should have some DEBUG flag (PHP var? GET parameter?) to make openflights swap between using the minified JS and the non minified versions where appropriate...

chrisrosset: Would including the source map be sufficient?

chrisrosset:
> Sorry for the various conflicts

How do I replicate your formatting updates? We need to re-do the formatting on this branch to make the merge possible - it's one gigantic conflict right now. If it's not something I can run easily, you could make the change yourself and push to my branch.

index.php
```diff
@@ -13,19 +13,18 @@
...
     <link rel="stylesheet" href="/css/style_reset.css" type="text/css">
     <link rel="stylesheet" href="<?php echo fileUrlWithDate("/openflights.css") ?>" type="text/css">
+    <link rel="stylesheet" type="text/css" href="/css/autocomplete.css">
     <link rel="gettext" type="application/x-po" href="/locale/<?php echo $locale;?>/LC_MESSAGES/messages.po?20090715" />
     <link rel="icon" type="image/png" href="/img/icon_favicon.png"/>
     <script type="text/javascript" src="https://www.google.com/jsapi"></script>
     <script type="text/javascript" src="/OpenLayers.js"></script>
     <script type="text/javascript" src="/js/greatcircle.js"></script>
-    <script type="text/javascript" src="/js/utilities.min.js"></script>
-    <script type="text/javascript" src="/js/Gettext.min.js"></script>
```
reedy, line 22: Looks like I managed to eat those on the merge...

js/autocomplete.js
reedy: I do get an error relating to the map being missing in the console - https://firefox-source-docs.mozilla.org/devtools-user/debugger/source_map_errors/#source-map-missing-or-inaccessible

reedy:
Force push was me undoing my merge as it was messy...

I've partially reverted 5f82d798e170414eee7b145cb254d8319b28fc84 in 3257782626bc54d4a51263b89239e86adacc4fed just to make life easier...

reedy:
Merging is a pain. Rebasing is much more helpful.

Fixed.

reedy:
Definitely seems a lot faster than the old one! :)

I think it's probably my fault, but the `Reg.` field has been forced down...

Screenshot 2023-07-09 at 21 43 49

index.php
```diff
@@ -227,7 +225,7 @@
 
   <td><?php echo _("Plane") ?></td>
   <td>
-      <input style="width: 180px" type="text" name="plane" id="plane" tabindex="111" onChange="JavaScript:markAsChanged();" onKeyDown='keyPress(event, "plane")' class="hintTextbox" value="<?php echo _("Enter plane model") ?>"><div id="planeAC"></div><input type="hidden" id="plane_id" name="plane_id" />
+      <input style="width: 180px" type="text" name="plane" id="plane" tabindex="111" onChange="JavaScript:markAsChanged();" onKeyDown='keyPress(event, "plane")' class="hintTextbox" value="<?php echo _("Enter plane model") ?>"><div id="planeAC"></div><input type="hidden" id="planeid" name="planeid" />
```
reedy: Ah, that div should've gone... Let me fix that

reedy: And that's fixed the display issue too!

reedy:
Screenshot 2023-07-09 at 21 49 06

Screenshot 2023-07-09 at 21 49 01

It's more something changing, rather than it being anything wrong... So "IT'S DIFFERENT" which is noticeable

Do we want to want to make the font, font size and colour the same as before? Do we care (I know you mentioned it in your original PR description)?

I think I find the font (and the size?) more readable.

The green stands out more to me, as it doesn't seem to match the colour palette of the site.

js/autocomplete.js
reedy: But I think the map is probably sufficient, rather than necessarily having the unminified version too.

Just need to resolve what we want to do about recording the version being used in the `git log` for ease of digging it up.

reedy:
```
Source map error: Error: request failed with status 404
Resource URL: https://openflights/js/autocomplete.js
Source Map URL: autocomplete.min.js.map
```

chrisrosset:
I wasn't sure this was going to get merged, so I focused on the
functionality and took the stock CSS. I don't know much CSS so there's that
too - though maybe all we need is to edit the color?

The old style is more consistent with the rest of the site.

css/autocomplete.css
```diff
@@ -0,0 +1 @@
+.autocomplete{background:#fff;z-index:1000;font:14px/22px "-apple-system",BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue",Arial,sans-serif;overflow:auto;box-sizing:border-box;border:1px solid rgba(50,50,50,.6)}.autocomplete *{font:inherit}.autocomplete>div{padding:0 4px}.autocomplete .group{background:#eee}.autocomplete>div.selected,.autocomplete>div:hover:not(.group){background:#81ca91;cursor:pointer}
```
reedy: `#81ca91` is the green

So if we override this in openflights.css... Let met test...

index.php
```diff
@@ -13,6 +13,7 @@
     <meta http-equiv="Content-Security-Policy" content="upgrade-insecure-requests; block-all-mixed-content;">
     <link rel="stylesheet" href="/css/style_reset.css" type="text/css">
     <link rel="stylesheet" href="<?php echo fileUrlWithDate("/openflights.css") ?>" type="text/css">
+    <link rel="stylesheet" type="text/css" href="/css/autocomplete.css">
```
reedy: Probably moving the order of this too...

reedy:
Screenshot 2023-07-09 at 22 04 32

Screenshot 2023-07-09 at 22 09 21

...

Old:
Screenshot 2023-07-09 at 21 49 06

New, green:
Screenshot 2023-07-09 at 21 49 01

New, old blue:
Screenshot 2023-07-09 at 22 08 32

Looks better to me, tbh :)

chrisrosset:
Looks great!

Would you like me to rebase/squash all of this into one commit (with the source map and the version in the commit message)?

Also, regarding the formatting, I have quite a few other branches that are affected... If we agree on a formatting tool (I don't care which one, I just need to consistently replicate whatever you want), then I can get those to work. Plus, you could also set up CI for formatting. :)

reedy:
> Would you like me to rebase/squash all of this into one commit (with the source map and the version in the commit message)?

Sounds good to me. Can merge this probably tomorrow (as I'm on my way to bed soon).

chrisrosset:
Cool, I'll make the changes later in the evening then.

jpatokal:
Whoah, thanks, this looks awesome!

One nit: I'm not a huge fan of the dark green highlight, can we keep the old light blue?

reedy:
> Whoah, thanks, this looks awesome!
> 
> 
> 
> One nit: I'm not a huge fan of the dark green highlight, can we keep the old light blue?

Yeah, I fixed that last night as the green stood out too much for me too, and wasn't in keeping with the colour palette of the site currently in use.

So it should be like this:

> New, old blue:
> 
> Screenshot 2023-07-09 at 22 08 32

reedy:
Does technically break the widget code (more than it may already be) by us removing `prototype.js`, but shouldn't matter because #1250 should be following in short order.

PR #1227, opened 2023-07-07
Open API documentation: login and logout

chrisrosset:
@jpatokal, @reedy As I'm making changes to the endpoints (to have them return JSON instead of their current bespoke serialization scheme; e.g. #1222), I think it would be nice to start documenting the available endpoints. This should make it easier to review them and to keep them consistent as they are developed (e.g. don't mix numbers/strings for the same value across requests, share object types between endpoints).

Doing this across the board is quite a bit of work so this PR provides minimal coverage for login/logout. If we agree this a good idea, we could build upon this.

You can preview this online by opening https://petstore.swagger.io/ and opening
https://raw.githubusercontent.com/chrisrosset/openflights/openapi-swagger/openapi.yaml
which links to the file on the branch for this PR.

This idea also came up in #876 (in [the linked Google Doc](https://docs.google.com/document/d/1yyNB-yDLVewyNpJ78ZvC3w6K3MB1NUvBFdNNFuET-9c/edit?disco=AAAAC45aHTU)).

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1,0 +1,126 @@
+openapi: 3.1.0
+info:
+  version: 0.0.1
+  title: OpenFlights Internal API Documentation
+externalDocs:
+  description: Project source code
+  url: https://github.com/jpatokal/openflights
+tags:
+  - name: login
+    description: User session management
+paths:
+  /php/login.php:
+    post:
+      tags:
+        - login
+      summary: Logs user into the system and creates a session
+      operationId: loginUser
+      parameters:
+        - name: challenge
+          in: query
+          description: Server-generated challenge string returned from the `php/map.php` endpoint
+          required: true
+          schema:
+            type: string
+        - name: name
+          in: query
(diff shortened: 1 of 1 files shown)


chrisrosset:
You can also use the Swagger UI to open [an enhanced version of the API doc with a proposal for a new "REST-inspired" `trip.php` endpoint](https://raw.githubusercontent.com/chrisrosset/openflights/rest-ish-trip/openapi.yaml) which I've created on a separate branch.

jpatokal:
I'm OK to merge this and have done so, but in general I am concerned about people mistakenly thinking that OpenFlights offers a public-facing API, because it does not and the existing private API is not architected to handle the myriad concerns that would bring up (rate limits, tokens, etc).  So let's be super clear about this in the docs.

reedy:
@chrisrosset FWIW, you should run prettier over this file, and/or ignore it in the `.prettierignore` :)

PR #1230, opened 2023-07-08
Swap while loops for foreach in DB calls

reedy:


php/apsearch.php
```diff
@@ -254,9 +255,10 @@
 if (!$sth->execute($params)) {
     die(json_encode(array("status" => 0, "message" => 'Operation ' . $param . ' failed.')));
 }
-while ($rows[] = $sth->fetch(PDO::FETCH_ASSOC));
+
+$rows = $sth->fetchAll(PDO::FETCH_NUM);
 array_pop($rows);
-foreach ($rows as &$row) {
```
reedy, line 259: Dropped the `&` here...

reedy (commented):
Needs fixing up

PR #1250, opened 2023-07-10
Remove the widget code

### LINKED ISSUE #1224: Proposal: remove widget code
@jpatokal, I would like to propose the removal of the map widget code.

The map widget hasn't seen any meaningful updates since it was first created in Feb 2010 (880b3f3b1b5ef396826aa72a224264f9397a23fe and 880b3f3b1b5ef396826aa72a224264f9397a23fe). I'm afraid it's beyond redemption - I doubt this code can be reasonably embedded into any modern website.

Removing this code would reduce the changes needed to support transitioning to JSON-based server APIs (which I've already started in #1222 with `top10.php`) and upgrading OpenLayers (ongoing discussion in #1192 with some work in progress). It would also allow us to remove `prototype.js` if the autocomplete upgrade (#1223) gets merged.

chrisrosset:
This code hasn't seen any action since it was created in early 2010. It has suffered major bitrot. The state of the code makes it impossible to actually integrate into any modern website. The fact that the widget requires running your own instance of OpenFlights is also a major dealbreaker.

The demo as seen on the actual website is broken in a number of ways: https://openflights.org/widget/demo

Closes https://github.com/jpatokal/openflights/issues/1224.


chrisrosset:
ping @reedy

reedy:
Best thing about version control is it can be reverted, restored or anything else in future if someone wants it.

reedy:
I'm presuming the pages don't really get much/any traffic... But either way, I'm wondering if we should leave some sort of index.html placeholder (could include rewrite rules, but maybe that's excessive) behind?

reedy:
> I'm presuming the pages don't really get much/any traffic... But either way, I'm wondering if we should leave some sort of index.html placeholder (could include rewrite rules, but maybe that's excessive) behind?

Meh, can be done in a followup if @jpatokal thinks it might be worth

Issue #1235, opened 2023-07-08
openflights.js: Inconsistent indenting


reedy:
Tabs, spaces... Not actually aligned...

chrisrosset:
What do you think about doing using something like `prettier` to auto-format all the JavaScript? We could also set up a PR hook to check formatting.

reedy:
Some sort of automation would be good.

I've not really touched any of the JS to this extent, but something we can wire into CI would definitely be the way forward.

There's numerous tools/libraries for enforcing JS (and CSS etc) et al coding standards.

Somewhere down the line we probably want to look at minification of the JS etc for performance in deployments.

Not that the repo has that much JS... it's all gonna help :)

chrisrosset:
Linking a related formatting discussion: https://github.com/jpatokal/openflights/pull/1223#issuecomment-1627829530

reedy:
> What do you think about doing using something like `prettier` to auto-format all the JavaScript?

I don't personally have any strong opinions either way...

https://eslint.org/ would be another option, as it does some amount of static analysis too

chrisrosset:
ESLint is not recommended for formatting these days. Best to keep formatting and linting separate.

Enabling ESLint would be useful but I would recommend silencing all the warnings at first and then enabling them one by one. There's going to be a lot of warnings. :)

I'll leave the formatting decisions to you since you've already started the work and you also have the privs to enable checks on PRs.

PR #1255, opened 2023-07-10
Prettier

reedy:
Ping #1235


chrisrosset:
If you're renaming `openlayers.js` maybe we could move it to `/js`?

js/jquery.blockUI.js
```diff
@@ -12,488 +12,547 @@
...
-	var t = ((p.offsetHeight - el.offsetHeight)/2) - sz(p,'borderTopWidth');
-	if (x) s.left = l > 0 ? (l+'px') : '0';
-	if (y) s.top  = t > 0 ? (t+'px') : '0';
-};
-
-function sz(el, p) {
-	return parseInt($.css(el,p))||0;
-};
-
+(function ($) {
```
chrisrosset: I think we should exclude this 3rd party file. We should separate all the 3rd party stuff out at some point.

reedy: Yeah, I was just looking around for a minified version of this...

Which lead to the rabbit hole of fixing the docs - https://github.com/malsup/blockui/pull/162

test/server/README.md
```diff
@@ -1,13 +1,12 @@
 This test suite validates PHP calls and MySQL database content.
 
-*** Instructions
+\*\*\* Instructions
```
chrisrosset: This looks weird - is this definitely right?

sql/README
```diff
@@ -1,23 +1,23 @@
-*** Tools for users
+\*\*\* Tools for users
```
chrisrosset: The changes in this file should be undone - it's plaintext, not markdown, and the indents were used on purpose.

test/server/README.md
reedy: The very opiniated prettier thinks so ;)

chrisrosset: I've never seen it do this. Let me take a quick look.

chrisrosset: Note that the main README.md file does not suffer the same "fix."

chrisrosset: OK, the original is wrong - I thought it was hashes.

```suggestion
### Instructions
```

sql/README
reedy: I wonder if we should make it MD for consistency...

reedy:
Dropped `effects.js` too as unused. Probably should've gone with eb830a82787939a1ca999b0b406da88f82dc8280

reedy:
Probably GTG at this point...

CI running correctly for direct push would be nice, but that can happen later...

sql/README.md
```diff
@@ -0,0 +1,14 @@
+### Tools for users
+
+- [`create.sql`](create.sql) - Create empty OpenFlights database schema
+- [`export.sql`](export.sql) - Export airlines.dat, airports.dat and routes.dat from OpenFlights database
```
reedy: This disappeared in 4ff8baf9d3daf43c8890cac9b08846a787fde420

reedy:
Any objects to merging this at this point?

chrisrosset:
> Any objects to merging this at this point?

This is great! Let's gooooooooooo.

PR #1259, opened 2023-07-10
Remove prototype.js

reedy:
Not sure why we cleaned it up in 8d20d44d5c625ed8fa9e545a9c035daf9aa01031...


reedy:
@chrisrosset This can go now, right? Especially as the widget code is gone too...

chrisrosset:
> @chrisrosset This can go now, right? Especially as the widget code is gone too...

I believe so. The only other reference to prototype I could find was in the OpenLayers code but I think they have their own copies of whatever they were using:

```
js/OpenLayers.min.js
16:/* Contains portions of Prototype.js:
```

Issue #1232, opened 2023-07-08
Make user ID for "demo mode" configurable


reedy:
Including Top 10 (logged out) usage etc

Forked from #1225

chrisrosset:
While we're here, I just wanted to point out another "special" User ID [defined in `config.php`](https://github.com/jpatokal/openflights/blob/52e054d87d63ba0d43153ff61ef8ed59b8fb0d3a/php/config.php.sample#L20).

I think it would be a good idea to define a similar constant to the demo ID and replace the hardcoded `1`s.

reedy:
I note `$OF_ADMIN_UID` isn't actually a constant, just a variable named like it could/should be one ;)

reedy:
> I note `$OF_ADMIN_UID` isn't actually a constant, just a variable named like it could/should be one ;)

And possibly, it should become an array too...

Doesn't seem unreasonable there'd be more than 1 "admin" account

chrisrosset:
> > I note `$OF_ADMIN_UID` isn't actually a constant, just a variable named like it could/should be one ;)
> 
> And possibly, it should become an array too...
> 
> Doesn't seem unreasonable there'd be more than 1 "admin" account

You're not wrong but I think it's unlikely there will be more than one. :)

PR #1267, opened 2023-07-10
Define and use $OF_DEMO_UID for demo mode

reedy:
Ref https://github.com/jpatokal/openflights/issues/1232

### DIFF
--- a/php/flights.php
+++ b/php/flights.php
@@ -43,17 +43,14 @@ if (!$uid || empty($uid)) {
     $uid = $_SESSION["openuid"];
     if ($uid && !empty($uid)) {
         // Yes we are, so check if we're limited to a single trip
-        $openTrid = $_SESSION["opentrid"];
-        if ($openTrid) {
-            if ($openTrid != $trid) {
-                // Naughty naughty, back to demo mode
-                $uid = 1;
-            }
+        if ($_SESSION["opentrid"] && $openTrid != $trid) {
+            // Naughty naughty, back to demo mode
+            $uid = $OF_DEMO_UID;
         }
         // No limit, do nothing
     } else {
         // Nope, default to demo mode
-        $uid = 1;
+        $uid = $OF_DEMO_UID;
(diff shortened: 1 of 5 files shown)


php/map.php
```diff
@@ -31,7 +31,7 @@
 $challenge = $_SESSION["challenge"] ?? null;
 if (!$uid || empty($uid)) {
     // If not logged in, default to demo mode and warn app that we're (no longer?) logged in
-    $uid = 1;
+    $uid = $_SESSION["opentrid"];
```
chrisrosset: What does a trip ID have to do with the user ID?

php/stats.php
```diff
@@ -8,7 +8,7 @@
 $public = "O"; // by default...
 
 // If not logged in, default to demo mode
-$uid = $_SESSION["uid"] ?? 1;
+$uid = $_SESSION["uid"] ?? $_SESSION["opentrid"];
```
chrisrosset: Same as above, expected `$OF_DEMO_UID` here.

php/map.php
reedy: I'm guessing I copied the wrong thing and wasn't paying any attention...

php/stats.php
reedy: Same here

php/flights.php
```diff
@@ -43,17 +43,14 @@
...
+            // Naughty naughty, back to demo mode
+            $uid = $OF_DEMO_UID;
         }
         // No limit, do nothing
     } else {
         // Nope, default to demo mode
-        $uid = 1;
+        $uid = $OF_DEMO_UID;
     }
 }
```
reedy, line 56: Some refactoring here would be nice to reduce the duplication a bit... But job for another day

reedy:
@jpatokal You'll probably want to pre-emptively set `$OF_DEMO_UID = 1;` in `config.php` in production before this change is deployed...

PR #1269, opened 2023-07-10
autocomplete.php: Fix some undefined array key errors

reedy:


reedy:
There's also some undefined variables too... `$query` on line 52 and line 132

chrisrosset:
I'm making some other changes to autocomplete.php right now. Let me fix these up there.

PR #1271, opened 2023-07-10
Implicitly declared variables

reedy:
I imagine there's a couple of false positives in here...

### DIFF
--- a/openflights.js
+++ b/openflights.js
@@ -128,6 +128,10 @@ var re_date =
 // Validate numeric value
 var re_numeric = /^[0-9]*$/;
 
+var lasturl;
+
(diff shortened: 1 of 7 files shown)


openflights.js
```diff
@@ -360,7 +364,7 @@ function init() {
     distance: 15,
     threshold: 3,
   });
-  airportLayer = new OpenLayers.Layer.Vector("Airports", {
+  var airportLayer = new OpenLayers.Layer.Vector("Airports", {
```
reedy: This one needs to be global

reedy:
Might be some more fallout, easily enough fixed though...

PR #1285, opened 2023-07-11
Assorted Autocomplete Fixes & Improvements

### LINKED ISSUE #1272: "XML Parsing Error: not well-formed" when using Carrier autocomplete
For some reason, typing in `b` or `B` results into the Carrier box when adding a new flight results in

```
XML Parsing Error: not well-formed
Location: https://openflights/
Line Number 1, Column 64:
```

### LINKED ISSUE #1283: Multi-input autocomplete popup size improvement
The multi-flight input has this feature (?) where the number of autocomplete results is reduced with each successive row.

The first row gets up to 6 suggestions, the second up to 5, ...

Relevant comment from `autocomplete.php`:

```
// Limit the number of rows returned in multi-input, where space is at a premium
```

This is done because the text fields are at the bottom of the screen.

But instead of doing all this work (this limits are worked out in a really funny way and passed all the way to the database), we could just pop the autocomplete window *above* the entry.

chrisrosset:
- Autocomplete display improvement for multi-input flight mode. Instead of decreasing the number of suggestions to control the size of the popup, we show it above the entry field. Fixes #1283.
- Fixed parameter encoding issues on the way to the server.
- Changed the parameter scheme for the endpoint to reduce coupling. The server no longer cares about the `id` of the `input` the search is for - it only cares about the type (airlines, airports, planes, mixed aka multisearch). Closes #1269.
- Converted server responses to JSON (fixes #1272).
- General refactoring of `autocomplete.php` to improve naming, fix warnings, simplify flow, etc.
- Moved some presentation logic to the UI.

I would love to remove the last autocomplete requests from `xmlHttpPost`... Those aren't really autocomplete as much as matching airline/airport codes and would probably be better done by making requests to the airport/airline search APIs but those are not that great either right now so I've just patched things for now. This behavior only happens when you tab out of the entry fields - but not if you change focus with the mouse... Would be great to remove that but then you'd have to "type, enter, tab" instead of "type, tab"... but maybe it's not that big of a deal?

### DIFF
--- a/openflights.css
+++ b/openflights.css
@@ -207,6 +207,14 @@ td.donate2 {
 }
 
 /* Autocomplete styles */
+.autocomplete {
+ /* TODO: improve this in code review
+  *
+  * We want the autocomplete suggestions to appear *over* the km/mi map
+  * distance measure when displaying them above the text entry. But how?
+  */
+  z-index: 9999;
+}
 .autocomplete > div.selected,
 .autocomplete > div:hover:not(.group) {
   background: #c6deff;
--- a/openflights.css
+++ b/openflights.css
@@ -208,10 +208,9 @@ td.donate2 {
 
 /* Autocomplete styles */
 .autocomplete {
- /* TODO: improve this in code review
-  *
+ /*
   * We want the autocomplete suggestions to appear *over* the km/mi map
-  * distance measure when displaying them above the text entry. But how?
+  * distance measure when displaying them above the text entry.
   */
   z-index: 9999;
 }
--- a/openflights.js
+++ b/openflights.js
@@ -409,7 +409,7 @@ function init() {
       query = arguments[1];
   }
 
-  prepareAutocomplete("qs", {
+  prepareAutocomplete("qs", "quick", {
     successCb: getQuickSearchId,
     failureCb: (e) => undefined,
   });
@@ -442,7 +442,7 @@ function init() {
     ac_plane = ["plane"];
 
     for (const airportAutoCompInputId of ac_airport) {
-      prepareAutocomplete(airportAutoCompInputId, {
+      prepareAutocomplete(airportAutoCompInputId, "airport", {
         successCb: (item) => {
           getSelectedApid(airportAutoCompInputId, item.value);
         },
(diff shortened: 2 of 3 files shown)


chrisrosset:
@reedy would be nice if this could go ahead #1271.

openflights.js
```diff
@@ -926,7 +957,15 @@ function xmlhttpPost(strURL, id, param) {
...
-        var cols = self.xmlHttpReq.responseText.split(";");
+        const resp = JSON.parse(self.xmlHttpReq.responseText);
+        var cols;
+        // Hack the old format back in (for now)
+        if (resp.length) {
+          cols = [resp[0].value, resp[0].label];
+        } else {
+          cols = [""];
+        }
+        console.log("cols", cols);
```
reedy: debug call

reedy:
Seems to work fine, and definitely fixes the various referenced fixed issues.

There's one `console.log()` left behind that was causing browser console spam

openflights.js
chrisrosset: removed

openflights.css
```diff
@@ -207,6 +207,14 @@ td.donate2 {
 }
 
 /* Autocomplete styles */
+.autocomplete {
+ /* TODO: improve this in code review
```
chrisrosset: @reedy Is there a better way to do this?

reedy:
I'm going to ask a silly question...

Where is the multi flight input? Is that part of the import?

reedy:
> Where is the multi flight input? Is that part of the import?

The "more" button on the basic screen... something I've never used

chrisrosset:
> > Where is the multi flight input? Is that part of the import?
> 
> The "more" button on the basic screen... something I've never used

The live version (at least for me) has a bug where one I fill the "To" on the 3rd row, the website hangs and maxes out the CPU.

Interestingly enough, I can't reproduce the problem. Maybe it was the autocomplete going haywire?

openflights.css
reedy: I'm not sure, but I also don't see a problem with this either...

You're working around a UI bug by setting a display order (of sorts), so setting it high it should show on top of basically everything...

reedy: It's also a temporary popup, not some UI element that is always on display

reedy:
What do you want to do about merging this?

I don't think we want to just squash it all into one (as there's some clearly distinct different changes), but there's a handful of commits (mostly the fixups/prettier etc) that could be squashed into other commits

chrisrosset:
I'll rebase a bit in a bit

reedy:
Thanks!

chrisrosset:
@reedy this is ready now

PR #1306, opened 2023-07-11
Allow admins to have premium features

### LINKED ISSUE #1279: Allow $OF_ADMIN_UID user(s) to have premium features
Following up #1232 and #1268

reedy:
Fixes https://github.com/jpatokal/openflights/issues/1279

### DIFF
--- a/html/settings.php
+++ b/html/settings.php
@@ -94,6 +94,11 @@ if ($type == "signup") {
     if (!$settings) {
         die(_("Could not load profile data"));
     }
+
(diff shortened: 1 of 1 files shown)


reedy:
I need to test this... I can't remember if this was enough (ie just to allow them to be set on settings)...

chrisrosset:
Some other places come to mind:

* `login.php` sets `$_SESSION['elite']`.
* `map.php` reads the DB and returns the elite status to the client. When opening a map for another user, this is used to display the right icon.

PR #1329, opened 2023-07-13
Exclude flights with unknown airline from unique airline count

### LINKED ISSUE #1328: Exclude "Unknown" from carrier count
Does anyone else feel that their airline count is wrong as a result of OpenFlights including "Unknown" as a carrier? I may have left the airline field blank on a flight because I can't remember whether it was operated by Air Canada or WestJet (both of which I've flown), but that doesn't make it a different airline.

I feel like this should be a fairly simple fix - if the user has any flights where the carrier is unknown, subtract 1 from the total - and if I knew the first thing about writing code, I'd propose a fix myself, but since this isn't Excel where I can just add a simple IF function, all I can do is make the suggestion 😃 

https://github.com/jpatokal/openflights/blob/290c84d00b49577577d6d96af27156940faf8f0a/openflights.js#L2274C1-L2277C20

reedy:
Fixes https://github.com/jpatokal/openflights/issues/1328

### DIFF
--- a/php/stats.php
+++ b/php/stats.php
@@ -74,8 +74,8 @@ if ($row) {
     $array += $row;
 }
 
-// unique airlines, unique planes, total distance (mi), average distance (localized), average duration
-$sql = "SELECT COUNT(DISTINCT alid) AS num_airlines, COUNT(DISTINCT plid) AS num_planes,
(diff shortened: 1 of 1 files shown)


chrisrosset:
DISTINCT DISTINCT?

reedy:
Copy paste fail 😂

PR #1337, opened 2023-07-14
Allow more strings to be localised

reedy:


reedy:
Reminder for myself... `json_error()` does `_($msg)`...

Which makes parameters in a string harder, and harder for xgettext to extract...

reedy:
I might need to start pushing these through (in some form, split or otherwise) to save rebase hell down the road ;)

reedy:
I think only b3dc81134bccebc9272ea0e5bb93e7e20f439147 and d7abb564170a6104f291480be0abcc10978a4399 left to come out from this branch...

PR #1339, opened 2023-07-14
Fix for Ghost Airport in Statistics

### LINKED ISSUE #993: Ghost Airport
When I do an analysis on my data set, FOA is reported as northernmost visited airport.
I have no record containing this airport in my data. northernmost should be LED

chrisrosset:
The SQL query used to list the northern/southern/eastern/western-most visited airports contains a logic error. As a result, the query may output airports the user has never visited.

The old query first identifies the maximum/minimum latitude/longitude and then looks up matching airports. This can produce invalid results for airports which share the same latitude or longitude with another airport.

The fix is to inverse the order. First, find all relevant flights and then find the geographical extremes.

The following query pulls up latitude duplicates and can be easily modified to pull up a similar list for longitude.

```
mysql> WITH dups AS (SELECT y, COUNT(*) as c FROM airports
    -> GROUP BY y HAVING c > 1)
    -> SELECT name, iata, y FROM airports
    -> WHERE y IN (SELECT y FROM dups) ORDER BY y;
+------------------------------------+------+---------------------+
| name                               | iata | y                   |
+------------------------------------+------+---------------------+
| Kerikeri Airport                   | KKE  |  -35.26279830932617 |
| Tumut Airport                      | TUM  |  -35.26279830932617 |
| Chinchilla Airport                 | CCL  | -26.774999618530273 |
| Fregon Airport                     | NULL | -26.774999618530273 |
| Kalkgurung Airport                 | KFG  | -17.431900024414062 |
| Mutoko Airport                     | NULL | -17.431900024414062 |
| Debre Tabor Airport                | DBT  |  11.967000007629395 |
| Obock Airport                      | OBC  |  11.967000007629395 |
| San Isidro Air Base                | NULL |  18.503700256347656 |
| Sangster International Airport     | MBJ  |  18.503700256347656 |
| Frank Pais International Airport   | HOG  |  20.785600662231445 |
| Lanai Airport                      | LNY  |  20.785600662231445 |
| Enshi Airport                      | ENH  |       30.3202991486 |
| Zagora Airport                     | OZG  |       30.3202991486 |
| Sfax Thyna International Airport   | SFA  |   34.71799850463867 |
| Paphos International Airport       | PFO  |   34.71799850463867 |
| Sivas Nuri Demira? Airport         | VAS  |           39.813801 |
| Rickenbacker International Airport | LCK  |           39.813801 |
| St Thomas Municipal Airport        | YQS  |   42.77000045776367 |
| Pamplona Airport                   | PNA  |   42.77000045776367 |
| RAF Wattisham                      | NULL |       52.1273002625 |
| Soesterberg Air Base               | UTC  |       52.1273002625 |
| Goose Bay Airport                  | YYR  |       53.3191986084 |
| Hinton/Jasper-Hinton Airport       | YJP  |       53.3191986084 |
| Gothenburg-Landvetter Airport      | GOT  |     57.662799835205 |
| Visby Airport                      | VBY  |     57.662799835205 |
| Pskov Airport                      | PKV  |   57.78390121459961 |
| Ruhnu Airfield                     | NULL |   57.78390121459961 |
| Stony Rapids Airport               | YSF  |  59.250301361083984 |
| Sanday Airport                     | NDY  |  59.250301361083984 |
+------------------------------------+------+---------------------+
30 rows in set (0.01 sec)
```

Fixes https://github.com/jpatokal/openflights/pull/993.

### DIFF
--- a/php/stats.php
+++ b/php/stats.php
@@ -135,67 +135,36 @@ foreach ($sth as $row) {
 }
 echo implode(";", $rows) . "\n";
 
-// North, South, West, East
-// 0 desc, 1 iata, 2 icao, 3 apid, 4 x, 5 y
 $sql = <<<SQL
-(
-    SELECT '%s', iata, icao, apid, x, y
-    FROM airports WHERE y = (
-        SELECT MAX(y)
-        FROM airports AS a, flights AS f
-        WHERE (f.src_apid = a.apid OR f.dst_apid = a.apid) AND $filter
-    )
-    ORDER BY iata
-    LIMIT 1
+WITH visited_airports AS (
+    SELECT a.iata, a.icao, a.apid, a.x, a.y
+    FROM flights f
+    JOIN airports a ON (f.src_apid = a.apid OR f.dst_apid = a.apid)
+    WHERE $filter
+    ORDER BY a.apid ASC
 )
-UNION
-(
-    SELECT '%s',iata,icao,apid,x,y
-    FROM airports
-    WHERE y = (
-        SELECT MIN(y)
-        FROM airports AS a, flights AS f
-        WHERE (f.src_apid = a.apid OR f.dst_apid = a.apid) AND $filter
-    )
-    ORDER BY iata
-    LIMIT 1
-)
-UNION
-(
-    SELECT '%s',iata,icao,apid,x,y
-    FROM airports
-    WHERE x = (
-        SELECT MIN(x)
-        FROM airports AS a, flights AS f
-        WHERE (f.src_apid = a.apid OR f.dst_apid = a.apid) AND $filter
-    )
-    ORDER BY iata
-    LIMIT 1
-)
-UNION
-(
-    SELECT '%s',iata,icao,apid,x,y
-    FROM airports WHERE x=(
-        SELECT MAX(x) FROM airports AS a, flights AS f WHERE (f.src_apid = a.apid OR f.dst_apid = a.apid) AND $filter
-    )
-    ORDER BY iata
-    LIMIT 1
-)
+(SELECT 'N' dir, a.* FROM visited_airports a ORDER BY y DESC LIMIT 1)
+UNION ALL
+(SELECT 'S' dir, a.* FROM visited_airports a ORDER BY y ASC LIMIT 1)
+UNION ALL
+(SELECT 'E' dir, a.* FROM visited_airports a ORDER BY x DESC LIMIT 1)
+UNION ALL
+(SELECT 'W' dir, a.* FROM visited_airports a ORDER BY x ASC LIMIT 1)
 SQL;
 
-$sth = $dbh->query(
-    sprintf(
-        $sql,
-        addslashes(_("Northernmost")),
-        addslashes(_("Southernmost")),
-        addslashes(_("Westernmost")),
-        addslashes(_("Easternmost"))
-    )
-);
+$compass_labels = [
+    'N' => _("Northernmost"),
+    'S' => _("Southernmost"),
+    'E' => _("Easternmost"),
+    'W' => _("Westernmost")
+];
+
+$sth = $dbh->query($sql);
 $rows = [];
 foreach ($sth as $row) {
-    $code = format_apcode2($row[1], $row[2]);
-    $rows[] = sprintf("%s,%s,%s,%s,%s", $row[0], $code, $row[3], $row[4], $row[5]);
+    $dir = $compass_labels[$row["dir"]];
+    $code = format_apcode2($row["iata"], $row["icao"]);
+    $rows[] = sprintf("%s,%s,%s,%s,%s", $dir, $code, $row["apid"], $row["x"], $row["y"]);
 }
 echo implode(":", $rows) . "\n";
 


php/stats.php
```diff
@@ -138,51 +138,29 @@
...
+    WHERE $filter
+    ORDER BY a.apid ASC
 )
+(SELECT 'N', a.* FROM visited_airports a ORDER BY y DESC LIMIT 1)
+UNION ALL
+(SELECT 'S', a.* FROM visited_airports a ORDER BY y ASC LIMIT 1)
+UNION ALL
+(SELECT 'W', a.* FROM visited_airports a ORDER BY x DESC LIMIT 1)
+UNION ALL
+(SELECT 'E', a.* FROM visited_airports a ORDER BY x ASC LIMIT 1)
```
reedy: I think there's something not quite right with the logic..

Before:

```
Northernmost | TRD (63.46°N 10.92°E)
Southernmost | ZQN (45.02°S 168.74°E)
Westernmost | PDX (45.59°N 122.6°W)
Easternmost | AKL (37.01°S 174.79°E)
```

After:

```
Airport records
Northernmost  TRD (63.46°N 10.92°E)
Southernmost  ZQN (45.02°S 168.74°E)
Westernmost AKL (37.01°S 174.79°E)
Easternmost PDX (45.59°N 122.6°W)
```

East and West look to be backwards.

Portland definitely isn't more east than Auckland :)

php/stats.php
```diff
@@ -194,8 +172,9 @@
 );
 $rows = [];
 foreach ($sth as $row) {
+    error_log(var_export($row, true));
```
reedy: Probably don't need the debugging leaving behind.

And then because github CR sucks, I can't leave comments on untouched lines...

You don't need to keep the insertion of the localised contexts into the SQL query anymore (not that there really was before either, just extra roundtripping).

```
$sth = $dbh->query(
    sprintf(
        $sql,
        addslashes(_("Northernmost")),
        addslashes(_("Southernmost")),
        addslashes(_("Westernmost")),
        addslashes(_("Easternmost"))
    )
);
```

should be fine as

`$sth = $dbh->query($sql);`

php/stats.php
```diff
@@ -138,51 +138,29 @@
...
-UNION
-(
-    SELECT '%s',iata,icao,apid,x,y
-    FROM airports WHERE x=(
-        SELECT MAX(x) FROM airports AS a, flights AS f WHERE (f.src_apid = a.apid OR f.dst_apid = a.apid) AND $filter
-    )
-    ORDER BY iata
-    LIMIT 1
+WITH visited_airports AS (
+    SELECT a.iata, a.apid, a.icao, a.x, a.y y
```
reedy: `a.y y` probably can just be `a.y`; the aliasing shouldn't be needed?

And the comment above the query is is now wrong

`// 0 desc, 1 iata, 2 icao, 3 apid, 4 x, 5 y`

The query output order was

```
+---+------+------+------+---------------+---------------------+
| N | iata | icao | apid | x             | y                   |
+---+------+------+------+---------------+---------------------+
```

But it's now

```
+---+------+------+------+---------------+---------------------+
| N | iata | apid | icao | x             | y                   |
+---+------+------+------+---------------+---------------------+
```

So the comment can become

`// 0 desc, 1 iata, 2 apid, 3 icao, 4 x, 5 y`

Or, you can reorder the params in `SELECT a.iata, a.apid, a.icao, a.x, a.y` to `SELECT a.iata, a.icao, a.apid, a.x, a.y` to match the old format.

Then the comment doesn't need updating!

php/stats.php
```diff
@@ -194,8 +172,9 @@
 );
 $rows = [];
 foreach ($sth as $row) {
+    error_log(var_export($row, true));
     $code = format_apcode2($row[1], $row[2]);
-    $rows[] = sprintf("%s,%s,%s,%s,%s", $row[0], $code, $row[3], $row[4], $row[5]);
+    $rows[] = sprintf("%s,%s,%s,%s,%s", $compass_labels[$row[0]], $code, $row[3], $row[4], $row[5]);
```
reedy: `function format_apcode2($iata, $icao) {`

`$code = format_apcode2($row[1], $row[2]);`
to
`$code = format_apcode2($row[1], $row[3]);`

because

```
+---+------+------+------+---------------+---------------------+
| N | iata | apid | icao | x             | y                   |
+---+------+------+------+---------------+---------------------+
```

`SELECT a.iata, a.apid, a.icao, a.x, a.y`
`// 0 desc, 1 iata, 2 apid, 3 icao, 4 x, 5 y`

---

Old:

```
+---+------+------+------+---------------+---------------------+
| N | iata | icao | apid | x             | y                   |
+---+------+------+------+---------------+---------------------+
```

`// 0 desc, 1 iata, 2 icao, 3 apid, 4 x, 5 y`
`$rows[] = sprintf("%s,%s,%s,%s,%s", $row[0], $code, $row[3], $row[4], $row[5]);`

So this was: `desc`, `$code`, `apid`, `x`, `y`.

New:

```
+---+------+------+------+---------------+---------------------+
| N | iata | apid | icao | x             | y                   |
+---+------+------+------+---------------+---------------------+
```

`SELECT a.iata, a.apid, a.icao, a.x, a.y`
`// 0 desc, 1 iata, 2 apid, 3 icao, 4 x, 5 y`
`$rows[] = sprintf("%s,%s,%s,%s,%s", $compass_labels[$row[0]], $code, $row[3], $row[4], $row[5]);`

Which outputs: `desc`, `$code`, `icao`, `x`, `y`.

Probably wants to be...

`$rows[] = sprintf("%s,%s,%s,%s,%s", $compass_labels[$row[0]], $code, $row[3], $row[4], $row[5]);`

To not change the "api output"...

Or as I mentioned above...

>Or, you can reorder the params in `SELECT a.iata, a.apid, a.icao, a.x, a.y` to `SELECT a.iata, a.icao, a.apid, a.x, a.y` to match the old format.

Then shouldn't need reordering here... either.

---

Which also fixes the broken hyperlinks for the airports (which currently causes some nice console spam)

Old:
`<a href="#" onclick="JavaScript:selectAirport(3720);">PDX</a>`

New:
`<a href="#" onclick="JavaScript:selectAirport(KPDX);">PDX</a>`

reedy (changes requested):

php/stats.php
reedy: I think

```
(SELECT 'W', a.* FROM visited_airports a ORDER BY x DESC LIMIT 1)
UNION ALL
(SELECT 'E', a.* FROM visited_airports a ORDER BY x ASC LIMIT 1)
```

into

```
(SELECT 'W', a.* FROM visited_airports a ORDER BY x ASC LIMIT 1)
UNION ALL
(SELECT 'E', a.* FROM visited_airports a ORDER BY x DESC LIMIT 1)
```

Should fix it...

The old query did `MAX`, `MIN`, `MIN`, `MAX` for N, S, W, E.

Where the new one is basically doing `MAX`, `MIN`, `MAX`, `MIN` for N, S, W, E.

--- 

Arguably, it should do it in N, S, E, W order as is more common... Rather than N, S, W, E.

So the other solution would be to just swap E and W around

```
(SELECT 'E', a.* FROM visited_airports a ORDER BY x DESC LIMIT 1)
UNION ALL
(SELECT 'W', a.* FROM visited_airports a ORDER BY x ASC LIMIT 1)
```

I'm just not sure offhand if that will require the JS to be updated...

reedy: Nope, looks like it doesn't care, just outputs them in the order it gets them

```
table +=
      "<tr><th colspan=2>" + gt.gettext("Airport records") + "</th></tr>";
    // TODO: Shouldn't be needed when stats.php returns full JSON
    // https://github.com/jpatokal/openflights/issues/1296
    var rows = extremes !== "" ? extremes.split(":") : [];
    for (r = 0; r < rows.length; r++) {
      var col = rows[r].split(",");
      // 0 desc, 1 code, 2 apid, 3 x, 4 y
      var lat = parseFloat(col[4]).toFixed(2);
      var lon = parseFloat(col[3]).toFixed(2);
      if (lat < 0) {
        lat = -lat + "&deg;S";
      } else {
        lat += "&deg;N";
      }
      if (lon < 0) {
        lon = -lon + "&deg;W";
      } else {
        lon += "&deg;E";
      }
      table +=
        "<tr><td>" +
        col[0] +
        '</td><td><a href="#" onclick="JavaScript:selectAirport(' +
        col[2] +
        ');">' +
        col[1] +
        "</a> (" +
        lat +
        " " +
        lon +
        ")</td></tr>";
    }
```

reedy: But note, again, the comment of `// 0 desc, 1 code, 2 apid, 3 x, 4 y` in openflights.js which is currently wrong/out of date etc :)

reedy (changes requested):

php/stats.php
reedy: It's an example where using indexing can cause problems or confusion.

Using column based (whether `$row['columnname']` or `$row-> columnname`) is often a lot clearer. Especially as you'll get errors if you typo, or don't select a column...

And you can reorder columns too without causing any issues!

php/stats.php
chrisrosset: Yeah, forgot to remove these.

reedy:
Funny how a simpler solution is a better one. Thanks!

PR #1345, opened 2023-07-15
Return JSON-encoded results from `stats.php`

### LINKED ISSUE #1296: stats.php should not return a mix of json and plain text
```
    var master = str.split("\n");
    var uniques = JSON.parse(master[0]);
    var longshort = master[1];
    var extremes = master[2];
    var classData = master[3];
    var reasonData = master[4];
    var seatData = master[5];
    var modeData = master[6];
    var classDataByDistance = master[7];
```

chrisrosset:
Fixes #1296.

### DIFF
--- a/openflights.js
+++ b/openflights.js
@@ -103,6 +103,8 @@ var classes,
   reasons,
   classes_short,
   reasons_short,
+  extreme_directions,
+  distance_units,
   modenames,
   modesegments,
   modeoperators,
@@ -196,6 +198,16 @@ function init() {
     T: gt.gettext("railway"),
     R: gt.gettext("road transport"),
     S: gt.gettext("shipping"),
+  };
+  extreme_directions = {
+    N: gt.gettext("Northernmost"),
+    S: gt.gettext("Southernmost"),
+    E: gt.gettext("Easternmost"),
+    W: gt.gettext("Westernmost"),
+  };
+  distance_units = {
+    km: gt.gettext("km"),
+    mi: gt.gettext("miles"),
   };
   topmodes = { F: gt.gettext("Segments"), D: gt.gettext("Mileage") };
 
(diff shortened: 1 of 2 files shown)


chrisrosset:
@reedy The locale-dependent number formatting changes conflict with my changes. I've resolved the conflicts with `-X ours` for now. Number formatting has to be re-done on this branch though to match the new expected behavior.

reedy:
> @reedy The locale-dependent number formatting changes conflict with my changes. I've resolved the conflicts with `-X ours` for now. Number formatting has to be re-done on this branch though to match the new expected behavior.

Arguably most of the formatting should probably be done in the frontend, but the current version is/was such a mix...

GitHub doesn't think they are resolved and there are currently conflicts...

openflights.js
```diff
@@ -2234,215 +2246,223 @@ function exportFlights(type, newWindow) {
...
+  link.setAttribute("href", "#");
+  link.setAttribute("onclick", `JavaScript:selectAirport(${apid});`);
+  link.append(text);
+  return link;
+};
+
+const formatCoordinates = ({ lat, lon }) => {
+  const latStr = Math.abs(lat).toFixed(2);
+  const lonStr = Math.abs(lon).toFixed(2);
+  return `${lonStr}°${lon >= 0 ? "N" : "S"} ${latStr}°${lat >= 0 ? "E" : "W"}`;
```
reedy, line 2260: Some localisation here in future I think :)

openflights.js
```diff
@@ -2234,215 +2246,223 @@ function exportFlights(type, newWindow) {
...
+    table += row.outerHTML;
+  }
+  table += "<tr><td>&nbsp;</td></tr>";
+
+  var distance = statsData.total.distance_mi;
+  table += "<tr><th colspan=2>" + gt.gettext("Distance") + "</th></tr>";
+  table +=
+    "<tr>" +
+    `<td>${gt.gettext("Total flown")}</td>` +
+    `<td>${statsData.total.distance} ${displayUnit}</td>` +
```
reedy: Formatting to possibly re-add?

openflights.js
```diff
@@ -2234,215 +2246,223 @@ function exportFlights(type, newWindow) {
...
+  ]) {
+    const flight = statsData.longshort[prop];
+    const row = document.createElement("tr");
+    const keyCol = document.createElement("td");
+    keyCol.append(gt.gettext(label));
+    const valCol = document.createElement("td");
+    valCol.append(createAirportLinkElement(flight.src_apid, flight.src_code));
+    valCol.append(" ↔ ");
+    valCol.append(createAirportLinkElement(flight.dst_apid, flight.dst_code));
+    valCol.append(`, ${flight.distance} ${displayUnit}, ${flight.duration}`);
```
reedy: Formatting to re-add to distance

openflights.js
```diff
@@ -2234,215 +2246,223 @@ function exportFlights(type, newWindow) {
...
+    row.append(valCol);
+    table += row.outerHTML;
   }
 
-  $("result").innerHTML = bigtable;
+  table +=
+    "<tr><td>" +
+    gt.gettext("Average") +
+    "</td><td>" +
+    statsData.average.distance +
```
reedy: Formatting to possibly re-add?

openflights.js
```diff
@@ -3842,7 +3860,7 @@ function login(str, param) {
           "<img src='/img/close.gif' height=17 width=17 onClick='JavaScript:closeNews()'> " +
           Gettext.strargs(
             gt.gettext(
-              "<B>Welcome to OpenFlights!</b> Click on %1 to start adding flights, or on %2 to load in existing flights from sites like FlightMemory."
+              " Click on %1 to start adding flights, or on %2 to load in existing flights from sites like FlightMemory."
```
reedy: leading space probably wants removing

php/stats.php
```diff
@@ -1,6 +1,8 @@
 <?php
 
-include_once 'locale.php';
+session_start();
+
+include_once 'config.php';
 include_once 'db_pdo.php';
 include_once 'helper.php';
 include_once 'filter.php';
```
reedy, line 7: We should clean up this loading... A little harder when it's mostly a load of global functions

php/stats.php
```diff
@@ -22,12 +24,12 @@
...
     $row = $sth->fetch();
     if (!$row) {
-        die('Error;' . _("Trip not found."));
+        die(json_encode(["error" => ("Trip not found.")]));
     }
 
     $public = $row["public"];
     if ($row["uid"] != $uid && $public == "N") {
-        die('Error;' . _("This trip is not public."));
+        die(json_encode(["error" => ("This trip is not public")]));
```
reedy: Keep the full stop for consistency? Or remove it from all of them?

Localistion dropped? `_(` has become `(`; brackets aren't adding anything in the current form

Could potentially use `json_error()` from `helper.php`.

php/stats.php
```diff
@@ -39,101 +41,121 @@
...
 $row = $sth->fetch();
 if ($row) {
-    $array += $row;
+    $response["unique"]["carriers"] = (int)$row["num_airlines"];
+    $response["unique"]["vehicles"] = (int)$row["num_planes"];
+    $response["total"]["segments"] = (int)$row["segments"];
+    $response["total"]["distance_mi"] = round((int)$row["distance"]);
+    $response["total"]["distance_km"] = round((int)$row["distance"] * KM_PER_MILE);
+    $response["total"]["distance"] = round((int)$row["distance"] * $unitMultiplier);
+    $response["total"]["duration_s"] = round((int)$row["distance"] * KM_PER_MILE);
```
reedy: Doing something like this, especially for multiple rows feels cleaner to me.. A variant on what you've done for some cases below.
```
$response["total"] += [
    "segments" => (int)$row["segments"],
    "distance_mi" => round((int)$row["distance"]),
    "distance_km" = round((int)$row["distance"] * KM_PER_MILE),
    "distance" => round((int)$row["distance"] * $unitMultiplier),
    "duration_s" => round((int)$row["distance"] * KM_PER_MILE),
];
```

php/stats.php
```diff
@@ -39,101 +41,121 @@
...
+    $response["average"]["distance"] = round((int)$row["avg_distance"] * $unitMultiplier);
+    $response["average"]["duration"] = $row["avg_duration"];
 }
-$array["avg_distance"] = number_format($array["avg_distance"]) . " " . $unit;
-$array["localedist"] = number_format(round($array["distance"] * ($units == "K" ? KM_PER_MILE : 1))) . " " . $unit;
-print json_encode($array) . "\n";
 
 // longest and shortest
 // 0 desc, 1 distance, 2 duration, 3 src_iata, 4 src_icao, 5 src_apid, 6 dst_iata, 7 dst_icao, 8 dst_apid
```
reedy: Does this still have value now we're not selecting by 0 based index?

php/stats.php
```diff
@@ -22,12 +24,12 @@
     $sth->execute([$trid]);
     $row = $sth->fetch();
```
reedy, line 24: https://www.php.net/manual/en/pdostatement.fetch.php

We should pass `PDO::FETCH_ASSOC` as the first parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -39,101 +41,121 @@
     $sth->execute([$user]);
     $row = $sth->fetch();
```
reedy, line 41: We should pass `PDO::FETCH_ASSOC` as the first parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -39,101 +41,121 @@
...
+    "by_mode" => [],
+    "by_class" => [],
+];
+
 // unique airports, and countries
 $sql = "SELECT COUNT(DISTINCT a.apid) AS num_airports, COUNT(DISTINCT a.country) AS num_countries
 FROM flights AS f,airports AS a
 WHERE (f.src_apid = a.apid OR f.dst_apid = a.apid) AND $filter";
 $sth = $dbh->query($sql);
 $row = $sth->fetch();
```
reedy, line 88: We should pass `PDO::FETCH_ASSOC` as the first parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -39,101 +41,121 @@
...
+       IFNULL(SUM(distance), 0) AS distance,
+       IFNULL(AVG(distance), 0) AS avg_distance,
+       IFNULL(SUM(TIME_TO_SEC(duration)), 0) AS duration,
+       IFNULL(TIME_FORMAT(SEC_TO_TIME(SUM(TIME_TO_SEC(duration))/COUNT(duration)), '%H:%i'), '00:00') AS avg_duration
 FROM flights AS f
-WHERE $filter";
+WHERE $filter
+SQL;
 $sth = $dbh->query($sql);
 $row = $sth->fetch();
```
reedy, line 109: We should pass `PDO::FETCH_ASSOC` as the first parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -39,101 +41,121 @@
...
-    sprintf(
-        $sql,
-        _("Longest"),
-        $multiplier,
-        _("Shortest"),
-        $multiplier
-    )
-);
-$rows = [];
+$sth = $dbh->query($sql);
```
reedy: https://www.php.net/manual/en/pdo.query.php

We should pass `PDO::FETCH_ASSOC` as the second parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -152,85 +174,89 @@
...
 SQL;
 
-$compass_labels = [
-    'N' => _("Northernmost"),
-    'S' => _("Southernmost"),
-    'E' => _("Easternmost"),
-    'W' => _("Westernmost")
-];
-
 $sth = $dbh->query($sql);
```
reedy: We should pass `PDO::FETCH_ASSOC` as the second parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -152,85 +174,89 @@
...
+// Classes
+$sql = <<<SQL
+    SELECT DISTINCT class, COUNT(*) segments, SUM(distance) distance
     FROM flights AS f
     WHERE $filter AND class != ''
     GROUP BY class
     ORDER BY class
-";
+SQL;
 $sth = $dbh->query($sql);
```
reedy: We should pass `PDO::FETCH_ASSOC` as the second parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -152,85 +174,89 @@
...
-$sql = "SELECT DISTINCT reason, COUNT(*)
+$sql = <<<SQL
+    SELECT DISTINCT reason, COUNT(*) segments, SUM(distance) distance
     FROM flights AS f
     WHERE $filter AND reason != ''
     GROUP BY reason
     ORDER BY reason
-";
+SQL;
 $sth = $dbh->query($sql);
```
reedy: We should pass `PDO::FETCH_ASSOC` as the second parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -152,85 +174,89 @@
...
-$sql = "SELECT DISTINCT seat_type, COUNT(*)
+$sql = <<<SQL
+    SELECT DISTINCT seat_type, COUNT(*) segments, SUM(distance) distance
     FROM flights AS f
     WHERE $filter AND seat_type != ''
     GROUP BY seat_type
     ORDER BY seat_type
-";
+SQL;
 $sth = $dbh->query($sql);
```
reedy: We should pass `PDO::FETCH_ASSOC` as the second parameter, as we don't need the 0 based indexed columns anymore.

php/stats.php
```diff
@@ -152,85 +174,89 @@
...
-$sql = "SELECT DISTINCT mode, COUNT(*)
+$sql = <<<SQL
+    SELECT DISTINCT mode, COUNT(*) segments, SUM(distance) distance
     FROM flights AS f
     WHERE $filter
     GROUP BY mode
     ORDER BY mode
-";
+SQL;
 $sth = $dbh->query($sql);
```
reedy: We should pass `PDO::FETCH_ASSOC` as the second parameter, as we don't need the 0 based indexed columns anymore.

reedy:
It seems to pull onto master fine...

openflights.js
```diff
@@ -2234,215 +2246,223 @@ function exportFlights(type, newWindow) {
...
-      gt.gettext("Reason") +
-      "</th><th>" +
-      gt.gettext("Seats") +
-      "</th></tr>";
-    table += "<tr>";
-    table += '<td><div id="chart_class"></div></td>';
-    table += '<td><div id="chart_reason"></div></td>';
-    table += '<td><div id="chart_seat"></div></td>';
-    table += "</tr>";
+  const displayUnit = gt.gettext(distance_units[statsData.distance_unit]);
```
reedy: The `gt.gettext()` call shouldn't be needed as the localised unit message is already in `distance_units[statsData.distance_unit]`?

openflights.js
```diff
@@ -2234,215 +2246,223 @@ function exportFlights(type, newWindow) {
...
+    ["Airports", "airports"],
+    ["Carriers", "carriers"],
+    ["Countries", "countries"],
+    ["Vehicles", "vehicles"],
+  ]) {
+    const row = document.createElement("tr");
+    const label = document.createElement("td");
+    label.append(gt.gettext(dataPoint[0]));
+    const value = document.createElement("td");
+    value.append(statsData.unique[dataPoint[1]]);
```
reedy: Formatting to possibly re-add?

php/stats.php
```diff
@@ -22,12 +24,12 @@
     $sth->execute([$trid]);
     $row = $sth->fetch();
     if (!$row) {
-        die('Error;' . _("Trip not found."));
+        die(json_encode(["error" => ("Trip not found.")]));
```
reedy: Localistion dropped? `_(` has become `(`; brackets aren't adding anything in the current form

Could potentially use `json_error()` from `helper.php`.

php/stats.php
```diff
@@ -39,101 +41,121 @@
     $sth->execute([$user]);
     $row = $sth->fetch();
     if (!$row) {
-        die('Error;' . _("User not found."));
+        die(json_encode(["error" => ("User not found.")]));
```
reedy: Localistion dropped? `_(` has become `(`; brackets aren't adding anything in the current form

Could potentially use `json_error()` from `helper.php`.

php/stats.php
```diff
@@ -39,101 +41,121 @@
...
     }
 
     $public = $row["public"];
     if ($public == "N") {
-        die('Error;' . _("This user's flights are not public."));
+        die(json_encode(["error" => ("This user's flights are not public.")]));
```
reedy: Localistion dropped? `_(` has become `(`; brackets aren't adding anything in the current form

Could potentially use `json_error()` from `helper.php`.

php/stats.php
```diff
@@ -39,101 +41,121 @@
...
 }
 $filter = "f.uid = " . $uid . getFilterString($dbh, $_POST);
-$array = [];
 
 // Convert mi to km if units=K
 $units = $_SESSION["units"] ?? null;
 if ($units == "K") {
-    $unit = _("km");
+    $unit = "km";
     $multiplier = "* " . KM_PER_MILE;
```
reedy: This doesn't seem to be used now.

php/stats.php
```diff
@@ -39,101 +41,121 @@
...
+    $unitMultiplier = KM_PER_MILE;
 } else {
-    $unit = _("mi");
+    $unit = "mi";
     $multiplier = "";
```
reedy: Same here

chrisrosset:
> It seems to pull onto master fine...

Yes, because the last commit here is a merge with master where I throw away (that's the `ours` above) any conflicting changes from master.

reedy (changes requested):
Seems to work fine! Few related changes we can make to reduce the amount of things being returned by the SQL queries etc

Comments about foramtting mostly left for my reference... I'm happy to rebase to include the lost changes in the merge commit when ther other minor issues have been fixed up.

openflights.js
chrisrosset: Yes, in the future. ;)

openflights.js
chrisrosset: I'm not sure where this change came from. I'm guessing I must have made this by accident. Will restore the old version.

php/stats.php
chrisrosset: Removed.

php/stats.php
chrisrosset: Removed.

php/stats.php
chrisrosset: Removed.

openflights.js
chrisrosset: Removed.

php/stats.php
chrisrosset: Done for all in this section.

php/stats.php
chrisrosset: I definitely need to do something here but I'm not trying to remove the translation. There is a `gettext` call for this string (for whatever is sent in the `error` prop in the UI now (line 2277). It works right now because we already have the string in the catalogue but the missing `_` is a problem. However, I don't want to do localization here in this file.

I definitely **don't** want to use `json_error` (and `json_success`) because, IMO, data endpoints should not send back localized messages. That's a presentation layer concern.

I'm thinking of defining `_` as the identity function in here so strings can be properly marked for localization in the UI.

reedy:
> > It seems to pull onto master fine...
> 
> Yes, because the last commit here is a merge with master where I throw away (that's the `ours` above) any conflicting changes from master.

I was referring to my previous comment

>GitHub doesn't think they are resolved and there are currently conflicts...

And is (still) showing

>This branch cannot be rebased due to conflicts

It might just be whatever git implementation github is using, sucking. I know jgit won't do trivial stuff that the "normal" git does fine/trivially with...

openflights.js
reedy: I might end up grabbing them in #1337 or a followup anyway

openflights.js
```diff
@@ -2234,215 +2246,223 @@ function exportFlights(type, newWindow) {
...
+  table += "<tr><th colspan=2>" + gt.gettext("Distance") + "</th></tr>";
+  table +=
+    "<tr>" +
+    `<td>${gt.gettext("Total flown")}</td>` +
+    `<td>${statsData.total.distance} ${displayUnit}</td>` +
+    "</tr>";
+  table +=
+    "<tr>" +
+    `<td>${gt.gettext("Around the world")}</td>` +
+    `<td>${(distance / EARTH_CIRCUMFERENCE).toFixed(2)}x</td>` +
```
reedy, line 2325: Another later TODO is possibly not having these defined different in both PHP and JS.... Possibly have PHP pass the constant value through when needed for calculations...

chrisrosset:
> > > It seems to pull onto master fine...
> > 
> > 
> > Yes, because the last commit here is a merge with master where I throw away (that's the `ours` above) any conflicting changes from master.
> 
> I was referring to my previous comment
> 
> > GitHub doesn't think they are resolved and there are currently conflicts...
> 
> And is (still) showing
> 
> > This branch cannot be rebased due to conflicts
> 
> It might just be whatever git implementation github is using, sucking. I know jgit won't do trivial stuff that the "normal" git does fine/trivially with...

Yeah, it can't be rebased which is why I merged master back here (throwing some changes away as a conflict resolution strategy). If you switch rebase to squash in the UI and you'll see it's happy to merge.

php/stats.php
reedy: It would be better having message codes, rather than strings passed wholesale to the localisation functions...

But I think that gets a little harder, changes how the defaults are defined (and found in source code) and a whole other host of problems.

So changes the workflow in one way or another too.

>I'm thinking of defining `_` as the identity function in here so strings can be properly marked for localization in the UI.

Which is what `locale.php` does for `if !$OF_USE_LOCALES`... And as long as no other PHP file includes `stats.php`, shouldn't cause an issue.

At some point, these want to become proper classes (especially if we go down the route of using a proper router etc), but that's a concern for down the road.

These could return an error code (and an english string? Ideally...), and then have the mapping from code => message in the JS and gettext should pick them up from there instead. And do the localisation there too.

And/or forcing a locale of `en_us` in standalone scripts like this, so the localisation function just returns the string anyway.

Not that `json_success`/`json_error` do any sort of code, just a status number...

reedy (commented):
So really only just the localisation string issue from the PHP files to resolve at this point.

Then squash the changes to something you're happy with.

I can then rebase and restore the formatting changes, and should be good to put onto master :)

Which should also make it easier for me to fix a last couple of cases (on a patch ontop) that were left behind too...

chrisrosset:
Made the `_` change and squashed on this branch as requested.

php/stats.php
```diff
@@ -1,10 +1,15 @@
...
-include_once 'locale.php';
+session_start();
+
+include_once 'config.php';
 include_once 'db_pdo.php';
 include_once 'helper.php';
 include_once 'filter.php';
 
+// Mark strings as localizable but don't localize.
+function _($str) { return $str; }
```
reedy: This actually causes breakages on my dev install (and therefore presumably would also in production).

```
"PHP message: PHP Fatal error:  Cannot redeclare _() in /var/www/openflights/php/stats.php on line 11
```

But I haven't explicitly installed the PHP gettext extension, I believe. Just gettext for the CLI commands...

```
# dpkg -S /usr/lib/php/20220829/gettext.so 
php8.2-common: /usr/lib/php/20220829/gettext.so
# dpkg -S /usr/lib/php/20190902/gettext.so 
php7.4-common: /usr/lib/php/20190902/gettext.so
```

And that's from the sury packages..

```
# dpkg -l | grep php | grep common
ii  php-common                            2:93+ubuntu22.04.1+deb.sury.org+2            all          Common files for PHP packages
ii  php7.4-common                         1:7.4.33-6+ubuntu22.04.1+deb.sury.org+1      amd64        documentation, examples and common module for PHP
ii  php8.2-common                         8.2.8-1+ubuntu22.04.1+deb.sury.org+1         amd64        documentation, examples and common module for PHP
```

I did set `$OF_USE_LOCALES = true;` to be able to test things.

If I set `$OF_USE_LOCALES = false;`, I get

```
PHP message: PHP Fatal error:  Cannot redeclare _() in /var/www/openflights/php/locale.php on line 28
```
Looking at my bash history, I don't seem to have explicitly installed it either.

But looking again at the INSTALL and #1206 neither explicitly install any of the `php-common` packages...

https://packages.ubuntu.com/jammy/amd64/php8.1 depends on https://packages.ubuntu.com/jammy/amd64/php8.1-common which depends on https://packages.ubuntu.com/jammy/amd64/php-common

```
# apt-cache --installed rdepends php7.4-common
php7.4-common
Reverse Depends:
  php7.4-zip
  php7.4-xml
  php7.4-readline
  php7.4-opcache
  php7.4-mysql
  php7.4-mbstring
  php7.4-json
  php7.4-gd
  php7.4-fpm
  php7.4-curl
  php7.4-cli
  php7.4
  php-common
# apt-cache --installed rdepends php8.2-common
php8.2-common
Reverse Depends:
  php8.2-zip
  php8.2-xml
  php8.2-xdebug
  php8.2-readline
  php8.2-opcache
  php8.2-mysql
  php8.2-mbstring
  php8.2-gd
  php8.2-fpm
  php8.2-curl
  php8.2-cli
# apt-cache --installed rdepends php-common
php-common
Reverse Depends:
  php8.2-zip
  php8.2-xml
  php8.2-xdebug
  php8.2-readline
  php8.2-opcache
  php8.2-mysql
  php8.2-mbstring
  php8.2-gd
  php8.2-curl
  php8.2-common
  php7.4-zip
  php7.4-xml
  php7.4-readline
  php7.4-opcache
  php7.4-mysql
  php7.4-mbstring
  php7.4-json
  php7.4-gd
  php7.4-curl
  php7.4-common
  php-zip
  php-xdebug
  php-mbstring
  php-json
  php-zip
  php-xdebug
  php-mbstring
  php-json
```

So you should have installed it on your Docker install too...

I'm guessing either something has changed with packaging (since whatever version the site was previously running on), as it's there in the default package on [22.04](https://packages.ubuntu.com/jammy/amd64/php8.1-common/filelist) (might need to refresh a few times, `packages.ubuntu.com` is giving intermittant HTTP 500)...

Does it work for you?

What do you have `$OF_USE_LOCALES` set to?

As far as I can tell... any calls to `php/stats.php` will be broken whichever way you've got it set...

And almost any other request (just loading the site) will be broken if you have `$OF_USE_LOCALES = false;` too...

reedy (changes requested):

php/stats.php
chrisrosset: Oof, yeah, it doesn't work. I admit I assumed it was fine based on `locale.php` already doing it. Ugh. And turning off locales breaks my local version as well.

reedy: I dunno if `$OF_USE_LOCALES` should be deprecated/removed, based on the existence of the apparent existence in default PHP...

Not stuff for this commit, but:

https://github.com/jpatokal/openflights/blob/5e0fae7fa7d44e4414d3ea28b585da1bc5c2712b/php/config.php.sample#L21 wants updating...

composer runs would actually fail if the "extension" wasn't loaded (unless you tried to use `--ignore-platform-reqs`) - https://github.com/jpatokal/openflights/blob/5e0fae7fa7d44e4414d3ea28b585da1bc5c2712b/composer.lock#L4346

https://github.com/jpatokal/openflights/blob/0f1adfc/php/locale.php#L28-L30 should either be removed (because of the composer requirement), or guarded with a `!function_exists('_')` call for it's definition (I doubt it's possible for `ext-gettext` to be apparently existing (ignoring `--ignore-platform-reqs`), but the function to not)...

Some distro's do do some weird stuff regarding packaging of php extensions... Some are included, some are pulled out etc...

But either way, both shouldn't be able to be true.

`$OF_USE_LOCALES` maybe/probably still has some value, basically unconditionally forcing `en_US` for all requests...

reedy: Heh, the `config.php.sample` comment is from c7b2cdf3967c8471eeddf284a45a038aba64278d (May 2013)...

reedy: Can you just set `$locale = "en_US.utf8";` at the top of the file?

Which, in theory, should "force" "en_US" for all `_()` calls in the file...

chrisrosset:
The `+=` change also seems to have broken the API response.

reedy:
> The `+=` change also seems to have broken the API response.

What's it complaining about?

It should work fine (we shouldn't have duplicate keys to worry about...), and shouldn't need to be using `array_merge()`...

reedy:
```
php > $response = [ "total" => [ "segments" => 0 ] ];
php > $response["total"] += [ "somethingelse" => "text" ];
php > var_dump( $response );
array(1) {
  ["total"]=>
  array(2) {
    ["segments"]=>
    int(0)
    ["somethingelse"]=>
    string(4) "text"
  }
}
```

php/stats.php
```diff
@@ -39,101 +44,126 @@
...
-$sth = $dbh->query($sql);
+WHERE $filter
+SQL;
+$sth = $dbh->query($sql, PDO::FETCH_ASSOC);
 $row = $sth->fetch();
 if ($row) {
-    $array += $row;
+    $response["unique"] += [
+        ["carriers"] => (int)$row["num_airlines"],
+        ["vehicles"] => (int)$row["num_planes"]
```
reedy: Don't want the brackets around these keys

php/stats.php
```diff
@@ -39,101 +44,126 @@
...
+    $response["total"] += [
+        "segments" => (int)$row["segments"],
+        "distance_mi" => round((int)$row["distance"]),
+        "distance_km" => round((int)$row["distance"] * KM_PER_MILE),
+        "distance" => round((int)$row["distance"] * $unitMultiplier),
+        "duration_s" => round((int)$row["distance"] * KM_PER_MILE),
+    ];
+    $response["average"] += [
+        ["distance"] => round((int)$row["avg_distance"] * $unitMultiplier),
+        ["duration"] => $row["avg_duration"]
```
reedy: Or here

reedy (changes requested):

reedy:
Won't like you trying to use an array as a key :)

chrisrosset:
I figured the `[]` problem first but then had to discover `+=` doesn't override values for existing keys... I set total segments at the very top because it allows the UI to easily determine we don't have any flights. Fixed it all now.

Still testing locale stuff now.

php/stats.php
chrisrosset: > Can you just set `$locale = "en_US.utf8";` at the top of the file?
> 
> Which, in theory, should "force" "en_US" for all `_()` calls in the file...

How would that work? Wouldn't `locale.php` override it with whatever is in `$_SESSION`?

chrisrosset:
I'm not sure what's going on but my PHP installation is not actually translating any strings right now. JS is fine...

reedy:
Anywhere? Or just this file?

You did re-add the include/require and drop the session start?

chrisrosset:
> Anywhere? Or just this file?
> 
> You did re-add the include/require and drop the session start?

Anywhere rendered server-side. 

I restored the `locale.php` include and removed `session_start`. I did run `locale-gen` before and all (and JS works). I think I must have assumed the language I chose had an incomplete translation.

php/stats.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 include_once 'locale.php';
+include_once 'config.php';
```
chrisrosset, line 4: We use `$OF_DEMO_UID` so we shouldn't rely on a transitive include from `locale.php`.

php/stats.php
```diff
@@ -22,12 +23,12 @@
     $sth->execute([$trid]);
```
chrisrosset, line 23: I can't comment on line 17 because I haven't changed it (I hate this) but I think the existing `$uid == 1` check is wrong... I think it's written with an assumption that you can only reach this from the demo screen.

chrisrosset:
OK, `gettext` the PHP extension needs enabling in `fpm/php.ini` apparently.

chrisrosset:
Alright, I'm not sure how to disable translation of the error messages but I'm also not super bothered about it happening... for now. It's also consistent with what I've done in `top10.php`. We should revisit our error-handling philosophy in general in the near future as we upgrade endpoints.

reedy:
https://stackoverflow.com/questions/24200471/php-gettext-without-relying-setlocale

chrisrosset:
> https://stackoverflow.com/questions/24200471/php-gettext-without-relying-setlocale

Not sure which part I'm supposed to look at? In any case, I think we're OK here for now.

reedy:
Just call `setlocale(LC_MESSAGES, "en_US");`

Shouldn't affect any other scope... or change session stuff

Issue #1352, opened 2023-07-15
GitHub url displayed in alert() box


reedy:
![Screenshot 2023-07-15 at 17 13 00](https://github.com/jpatokal/openflights/assets/67615/2b498f47-8884-4822-8fe4-42f8d7778ae0)

You can copy from the alert box, but it's not the most useable... And then the window immediately closes.

Should keep it open (it does update the text at the bottom before it closes) and probably make it a clickable link...

And maybe some nice colouring to suggest it worked ;)

PR #1354, opened 2023-07-15
Further GitHub issue improvements

### LINKED ISSUE #354: Distinguish new airports and edits to existing airports in the issues

### LINKED ISSUE #1318: GitHub search is too simplistic
```
            $issues = $github->api('search')->issues("repo:$GITHUB_USER/$GITHUB_REPO is:issue in:title $identifier");

            if (count($issues['items']) > 0) {
                // Existing issue, add comment
                $issueNumber = $issues['items'][0]['number'];
                // https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#create-an-issue-comment
                $result = $github->api('issue')->comments()->create(
                    $GITHUB_USER,
                    $GITHUB_REPO,
                    $issueNumber,
                    ['body' => $body]
                );
            } else {
                // New issue
                // https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#create-an-issue
                $result = $github->api('issue')->create(
                    $GITHUB_USER,
                    $GITHUB_REPO,
                    ['title' => $subject, 'body' => $body, 'labels' => ['airport']]
                );
                $issueNumber = $result['number'];
            }
```

This looks for any task with the identifier in the title... Which results in comments being left on unrelated tasks, like https://github.com/jpatokal/openflights/issues/499#issuecomment-203581482

reedy:
Further GitHub issue improvements

- Issue subject now varies if it's an addition or an edit
   - Fixes https://github.com/jpatokal/openflights/issues/354 by changing the subject if it's an addition or an edit
   - Also varies success message displayed to user
- Search for more specific issue subject
   - Looking specifically for an open issue, perferably with label:airport
   - First look for the the exact subject we'd be creating
   - Then looking for the other subject (which may be the old style, which was just "update")
   - And then look for the "($iata/$icao)" string, optionally with either missing, incase someone submitted an issue without on before
   - And stops looking for any task with the IATA/ICAO anywhere in the title (many false positives)
   - Will add label:airport to an existing (open) task if it doesn't have it already
   - Fixes https://github.com/jpatokal/openflights/issues/1318
- Improve issue/comment body
   - Only output possible duplicate list if there's any
   - Output cross search strings specifically for icao and iata if they are not empty strings
   - Output a search for the airport name
- Some code reordering so we only do some setup if we need it
   - Explicit exits so we can unindent some large blocks too

### DIFF
--- a/php/apsearch.php
+++ b/php/apsearch.php
@@ -1,5 +1,9 @@
 <?php
 
+use Github\Api\Issue;
+use Github\Api\Search;
+use Github\Client;
+
 require_once '../vendor/autoload.php';
 require_once '../php/locale.php';
 require_once '../php/db_pdo.php';
@@ -30,13 +34,14 @@ if ($action == "RECORD") {
     if (!$uid || empty($uid)) {
         json_error("Your session has timed out, please log in again.");
     }
+    $addAirport = !$apid || $apid == "";
 
     // Check for potential duplicates (unless admin)
     $duplicates = [];
     if (!in_array($uid, (array)$OF_ADMIN_UID)) {
         $filters = [];
         $filterParams = [];
-        if ($apid && $apid != "") {
+        if (!$addAirport) {
             $filters[] = "apid = ?";
             $filterParams[] = $apid;
         }
@@ -49,7 +54,11 @@ if ($action == "RECORD") {
             $filterParams[] = $icao;
         }
 
-        $sql = "SELECT * FROM airports WHERE " . implode(" OR ", $filters);
+        // Column order to match $newEdit below
+        $sql = "SELECT apid, name, city, country, iata, icao, x, y, elevation, timezone,
+            dst, country_code, uid, tz_id, type, source
+            FROM airports
+            WHERE " . implode(" OR ", $filters);
         $sth = $dbh->prepare($sql);
         $sth->execute($filterParams);
         foreach ($sth->fetchAll(PDO::FETCH_ASSOC) as $row) {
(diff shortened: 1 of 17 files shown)


php/apsearch.php
```diff
@@ -63,146 +64,265 @@
...
+
+TXT;
+}
 
-            $issues = $github->api('search')->issues("repo:$GITHUB_USER/$GITHUB_REPO is:issue in:title $identifier");
+    $body .= <<<TXT
+Name:
+- OurAirports: https://ourairports.com/search?q=$airport
+- Wikipedia: https://www.google.com/search?q=wikipedia%20$airport%20airport&btnI
+TXT;
```
reedy, line 230: html encoding?

php/apsearch.php
```diff
@@ -63,146 +64,265 @@
...
+        $github->authenticate($GITHUB_ACCESS_TOKEN, null, Github\AuthMethod::ACCESS_TOKEN);
+
+        // This can/will leave comments on closed issues...
+        // This could be because the newer is a dupe and may have been closed...
+        // Or it was declined...
+        $issues = [];
+        foreach ($searchStrings as $searchText) {
+            // https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-issues-and-pull-requests
+            $issues = $github->api('search')->issues(
+                "repo:$GITHUB_USER/$GITHUB_REPO is:issue in:title \"$searchText\""
```
chrisrosset: I think we should add `is:open` and `label:airport`.

php/apsearch.php
reedy: yup

reedy:
Doesn't fix #1352, but leaves a TODO about it...

PR #1360, opened 2023-07-17
Config setting for enabling/disabling signups

chrisrosset:
For running a private instance of OpenFlights which does not want to offer public access to the service prevent limit abuse.

### DIFF
--- a/html/settings.php
+++ b/html/settings.php
@@ -1,10 +1,16 @@
 <?php
 
+require_once "../php/config.php";
 require_once "../php/locale.php";
 require_once "../php/db_pdo.php";
 require_once "../php/helper.php";
 
 $type = isset($_GET["new"]) ? "signup" : "settings";
+
+if (isset($OF_ENABLE_SIGNUP) && !$OF_ENABLE_SIGNUP) {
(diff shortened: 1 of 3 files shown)


html/settings.php
```diff
@@ -1,11 +1,17 @@
 <?php
 
+require_once "../php/config.php";
 require_once "../php/locale.php";
 require_once "../php/db_pdo.php";
 require_once "../php/helper.php";
 
 $type = isset($_GET["new"]) ? "signup" : "settings";
 
+if (isset($OF_ENABLE_SIGNUP) && !$OF_ENABLE_SIGNUP) {
```
reedy, line 10: This reminds me of register_globals.

Definitely feels like we should be loading in a "default" config before loading config.php ontop.

Doesn't need fixing here though.

chrisrosset: Yeah, that would be nice. I went with the `isset` approach for backwards-compatibility but I think we've already broken it with #1267.

A configuration module/class would allow defining more than just variables/constants. That would be the only thing reading `config.php` and we could deal with compatibility there.

Issue #1376, opened 2023-07-18
Encode `flights.php` responses using JSON


chrisrosset:

reedy:
Makes #1076 a lot easier/basically does most of it

PR #1382, opened 2023-07-18
Move GreatCircleMapper link logic to the front-end

chrisrosset:
We can generate the GCMap URL based on the data we've just received from `flights.php` instead of sending another request to get redirected.

This removes one of the many different types of responses `flights.php` can return to its client. Might help with the future JSON conversion.

### DIFF
--- a/openflights.js
+++ b/openflights.js
@@ -2119,6 +2119,9 @@ function listFlights(str, desc, id) {
     desc = gt.gettext("Flights:") + " " + getMapTitle(false);
   }
 
+  // TODO: Won't be needed once the endpoint returns JSON.
+  const gcmFlightList = [];
+
   // IE string concat is painfully slow, so we use an array and join it instead
   var table = [];
   table.push(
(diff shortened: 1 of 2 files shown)


chrisrosset:
link #1376

php/flights.php
```diff
@@ -257,17 +252,4 @@
...
-
-if (count($rows)) {
-    $separator = ($export == "export" || $export == "backup")
-        ? "\r\n"
-        // Not for $export == "gcmap"
-        : "\n";
-
-    echo implode($separator, $rows);
-}
+echo implode("\r\n", $rows);
```
reedy: Not that there really should be much difference for most (decent) applications of `\r\n` vs `\n`, but...

https://github.com/jpatokal/openflights/commit/0f1adfcaae5d35ddb7337761da480d4fd34bec77#diff-21a477194502a985c437c81d564ce284eaff75d08a36e2d7e6ad6a72f017bcd6L175-L179

Do we pass anything else as a parameter value to `export=` other than `export` or `backup` after this patch?

reedy:
Looks good, will try and get it tested later today.

php/flights.php
chrisrosset: I don't think we pass other values to `export` but we do handle requests that don't set `export` (for "list flights"). I didn't see any breakage but the `\r` (we split on newlines in JS) would be invisible in the browser.

I'll revert this part of the change. This part will have to change anyway for us to send JSON back to the UI.

reedy: https://github.com/jpatokal/openflights/blame/899b1c53a53b1094448d12bf202092898e8024a9/php/flights.php#L181-L186

https://github.com/jpatokal/openflights/commit/cc8dfd79bfd6a83795b43bc0f9c1c69d8b7f7d55#diff-8c9cf639f05f3287fbb6ef307938f6a66297a4081b8eafbadc996dee50315fafR865

```
var rows = str.split("\n");
```

Yeah, I still see 3 `str.split("\n")`; `updateFilter`, `updateMap` and `listFlights` in openflights.js now.

Leaving trailing `\r` isn't the best, and like you say, when #1376 happens, this gets fixed properly anyway, so we don't need to worry about it too much!

reedy: Fixed in the rebase/force push.

openflights.js
```diff
@@ -2286,10 +2291,21 @@ function listFlights(str, desc, id) {
...
+
+  // Great Circle Mapper button setup
+  document.getElementById("gcmapbutton").onclick = () => {
+    const gcmURL = new URL("http://www.gcmap.com/mapui");
+    gcmURL.searchParams.append(
+      "P",
+      gcmFlightList.map((f) => `${f.src_code}-${f.dst_code}`).join(",")
+    );
+    gcmURL.searchParams.append("MS", "bm"); // 'M'ap 'S'tyle: 'b'lue 'm'arble
+    location.assign(gcmURL.href);
```
reedy: Can we open this in a new tab, to keep the same behaviour?

`window.open(gcmURL.href, "_blank");` works... Not sure if there's a nicer/better way offhand.

reedy (commented):
There's also #1389 causing some console logspam, but that's not caused by this; it's an existing issue, so don't need to worry about it.

Seems to work fine, just one minor change requested to keep the same behaviour.

openflights.js
reedy: Seems `window.open(url, "openflights_export");` is what we're already in the existing code

PR #1383, opened 2023-07-18
Add language selectors to most help pages

reedy:


reedy:
Need to actually test this some...

reedy:
"Language" seems to change fine for translated languages...

Not seeing the changes from the en_GB locale though...

PR #1388, opened 2023-07-19
Locale setup improvements

### LINKED ISSUE #1322: Select an appropriate locale from request
From a mixture of the `Accept-Language` header, and [`Navigator.languages`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages) we can attempt to change the UI of openflights (on initial load, if not already defined in preferences/similar) to something appropriate for the user

reedy:
- Check if $_GET["lang"] actually is a locale we can use
   - ?lang=foo gets noisy
- Try and set an appropriate locale from the ACCEPT_LANGUAGE header
- Push chosen $locale back into $_SESSION["locale"]
- Requires ext-intl now

Fixes https://github.com/jpatokal/openflights/issues/1322


Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN apt-get update && apt-get install -y \
 # System-level PHP setup
 RUN add-apt-repository ppa:ondrej/php -y
 RUN apt-get install -y \
-    php7.4 php7.4-curl php7.4-fpm php7.4-gd php7.4-mbstring php7.4-mysql php7.4-xml
+    php7.4 php7.4-curl php7.4-fpm php7.4-gd php7.4-intl php7.4-mbstring php7.4-mysql php7.4-xml
```
chrisrosset, line 21: Added the "server" label because of the new dependency.

php/locale.php
```diff
@@ -5,12 +5,37 @@
...
+    $locale = null;
     if (isset($_GET["lang"])) {
-        $locale = $_GET["lang"];
-        $_SESSION["locale"] = $locale;
-    } else {
-        $locale = $_SESSION["locale"] ?? "en_US";
+        $lang = $_GET["lang"];
+
+        // We're accepting a random lang parameter... We should probably check it actually exists!
+        if (locale_exists($dbh, $lang)) {
```
chrisrosset, line 15: Where was `$dbh` defined? We haven't included `pdo_db.php`.

php/locale.php
```diff
@@ -5,12 +5,37 @@
...
+        // If we've not already got a locale set (from $_SESSION or $_GET),
+        // see if we can use the HTTP ACCEPT_LANGUAGE header value
+        // TODO: We don't attempt any of the fallbacks...
+        $acceptLang = Locale::acceptFromHttp($_SERVER['HTTP_ACCEPT_LANGUAGE']);
+        // We know we can do en_US, so don't do a DB lookup
+        if ($acceptLang !== "en_US" && locale_exists($dbh, $acceptLang)) {
+            $locale = $acceptLang;
+        } else {
+            // Fallback to en_US from a lack of another/better option
+            $locale = "en_US";
```
chrisrosset: Couldn't this just be:
```php
if ($acceptLang === "en_US" || locale_exists($dbh, $acceptLang)) {
    $locale = $acceptLang;
}
```

reedy: Yup; the joys of coding when tired :)

Dockerfile
reedy: I'm thinking we should start potentially creating "deployment" tasks with an easy list for Jani of stuff that needs doing/changing since the last one.

php/locale.php
reedy: Yeah, it's not. Didn't test this well!

reedy:
I note this also fixes the issue that if you do `?lang=foo`, which spams you with errors, and then get stuck in your session until you do `?lang=en_US` or similar to make it sensible again

Screenshot 2023-07-20 at 11 35 41

Dockerfile
reedy: #1391

PR #1450, opened 2023-11-05
Fix typo (6 => 16) in csv.php.

dyung:
The page help/csv.php appears to have a typo (a missing 1) on it. In the description of the CSV fields, the OID first field is numbered 6, when it probably should be 16 given the context. This change fixes that.


reedy:
Thanks! Broken by 6ecdade7f8b4270d739fe5f15886af07870a506f

